PR #19, opened 2017-06-09
Prevent buffer overflow for mount point path

pierreN:
… in dlt_logstorage_open_log_file.

When using dlt-logstorage-ctrl utility, dlt_logstorage_open_log_file function calls sprintf with raw input (dlt-logstorage-ctrl -p argument) into a buffer on the stack (of size 50).

When the argument length is longer than 50, a buffer overflow happens.
I guess this could hypothetically create a small security issue if the user can choose himself the storage mount point path (depends of DLT usage).

This PR simply uses snprintf instead of sprintf and return an error if the stack variable is not big enough.

Signed-off-by: Pierre N <pierreN@users.noreply.github.com>


pierreN:
ping

PR #21, opened 2017-08-01
Data stuck in receiver buffer when dlt_daemon_user_send_log_level() fails

lhelwing:
- When dlt_daemon_user_send_log_level() called in
dlt_daemon_process_user_message_register_context() fails -1 was returned
which caused dlt_daemon_process_user_messages() to stop processing
receiver buffer. Remaining data was stuck until new data arrived over
FIFO.

- Make debug output of dlt_daemon_user_send_log_level() more
verbose.

Signed-off-by: Lutz Helwing <lutz_helwing@mentor.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,6 @@ config.h
 include/dlt/dlt_version.h
 .project
 .cproject
-
-
 .settings
+.idea/
+cmake-build-debug/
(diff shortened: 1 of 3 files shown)


lhelwing:
See mailing list thread "DLT function interface shows strange behaviour" for discussion

lhelwing:
I'm registering the context with:

DLT_REGISTER_CONTEXT_LL_TS()

Why is the daemon trying to send a log level at all? I thought when
DLT_REGISTER_CONTEXT_LL_TS the log level is set by the application
itself and thus the daemon must not send one. Or am i misunderstanding
something?

lhelwing:
What side effects could be expected when removing the "return -1"? Why does this error abort processing the read data?

PR #23, opened 2017-08-08
Porting CDltWrapper from audio manager to dlt-daemon

zeerd:
The class is useful for c++ users.
I port it here, deleted some codes in NO_DLT branch and add a test code.

Signed-off-by: Emneg Zeerd <emneg.zeerd@gmail.com>


clipka:
Why another C++ wrapper is needed? Could this be aligned with the existing one?

PR #26, opened 2017-08-30
Prevention for occasional corrupted messages 

baduras:
This patch prevents occasional corrupt messages, which occur mostly in situations, where system is highly utilized: e.g. socket send buffer is full (or situation when write function is interrupted).

### DIFF
--- a/src/daemon/dlt_daemon_socket.c
+++ b/src/daemon/dlt_daemon_socket.c
@@ -145,27 +145,81 @@ int dlt_daemon_socket_close(int sock)
 int dlt_daemon_socket_send(int sock,void* data1,int size1,void* data2,int size2,char serialheader)
 {
     /* Optional: Send serial header, if requested */
+    int ret = DLT_RETURN_OK;
     if (serialheader)
     {
-        if ( 0 > send(sock, dltSerialHeader,sizeof(dltSerialHeader),0) )
-            return DLT_DAEMON_ERROR_SEND_FAILED;
-
+         ret = dlt_daemon_socket_sendreliable(sock, (void *) dltSerialHeader, sizeof(dltSerialHeader));
+         if(ret != DLT_RETURN_OK)
+             return ret;
     }
 
     /* Send data */
-    if(data1 && size1>0)
+    if(data1 != NULL && size1 > 0)
     {
-        if (0 > send(sock, data1,size1,0))
-            return DLT_DAEMON_ERROR_SEND_FAILED;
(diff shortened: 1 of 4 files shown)


src/daemon/dlt_daemon_socket.c
```diff
@@ -145,27 +145,81 @@ int dlt_daemon_socket_close(int sock)
...
+   while (remaining_size > 0 && resend_ou <= DLT_DAEMON_SOCKET_MAXRESEND_ATTEMPTS)
+   {
+       int resend_in = 0;
+       ssize_t ret = 0;
+
+       //repeat due to some temporal issues
+       do
+       {
+            ret = send(sock, buffer + data_sent, remaining_size, 0);
+            if((ret < 0) && (errno != EINTR && errno != EAGAIN && errno != EWOULDBLOCK))
```
ManikandanChockalingam: Condition in line no:192 and 199 contradicts. retry print was printed if errno is none of the mentioned errors but actually retry was happening only on any of these errors

src/daemon/dlt_daemon_socket.c
```diff
@@ -145,27 +145,81 @@ int dlt_daemon_socket_close(int sock)
...
+            {
+               snprintf(str,DLT_DAEMON_TEXTBUFSIZE,"dlt_daemon_socket_sendreliable: socket send failed, but it is repeated again [errno: %d]!\n", errno);
+               dlt_log(LOG_WARNING,str);
```
ManikandanChockalingam: instead of snprintf and dlt_log, dlt_vlog can used

src/daemon/dlt_daemon_socket.c
```diff
@@ -145,27 +145,81 @@ int dlt_daemon_socket_close(int sock)
...
+           dlt_log(LOG_WARNING,str);
+           return DLT_DAEMON_ERROR_SEND_FAILED;
+       }
+
+       remaining_size -= ret;
+       data_sent += ret;
+
+       //there is data remaining, which should be resent, because corrupted messages may occur
+       if( remaining_size > 0 )
+       {
```
ManikandanChockalingam: codding style mismatch line no 201 and line no 212. better to follow unique style throughout.

src/daemon/dlt_daemon_socket.c
baduras: I agree, the message should be printed in case of given conditions occurred. 
Must be updated.

PR #29, opened 2017-09-15
remove linux from limits.h include

razr:


src/console/dlt-control.c
```diff
@@ -488,6 +488,7 @@ int main(int argc, char* argv[])
             return -1;
         }
     }
```
ManikandanChockalingam: define based check is not added in all the places where serial is used in dlt-control. In other test applications it is added but it is missed here. please update.

include/dlt/dlt_common.h
```diff
@@ -83,7 +83,7 @@
 #include <time.h>
 #endif
 
-#if !defined (__WIN32__) && !defined(_MSC_VER)
+#if defined (DLT_SERIAL)
 #include <termios.h>
```
ManikandanChockalingam: Why __WIN32__ and __MSC_VER  is replaced by DLT_SERIAL in many places? what is the relation between these ?

gunnarx:
Hello @razr Do you have any answers?  (I'm also curious to know... :-)

My feedback would be:
1. The commit comment "remove linux from limits.h include" explains what, but not why (put it in the text below). What's the problem with linux/limits?  Are you aiming to support compilation on other platform, and does this solve that issue?

In the PR itself, could you explain what you are trying to achieve (because removing the __WIN32__ is as @ManikandanChockalingam says also not explained.

2.  The two commits seem to be unrelated changes and ought to be separate PRs.

What's needed to drive this to a close? @clipka - do you plan to comment on these (individually)? Do you see a need for Linux-specific code in DLT or not - it's supposed to support Windows I assume?

Git blame tells me linux/limits.h was put in during a major feature change: https://github.com/GENIVI/dlt-daemon/commit/ed412acac8e9f8a5745e55adeed9652b34b9ee4f
by @lhelwing and stefan.vacek@intel.com.  - Comments?  What was the intention of **linux**/limits.h

razr:
I apologize for the late reply.
1. limit.h - not all OSes are Linux, there no need to use word linux in this particular include.

Two commits are unrelated indeed, I can resubmit them

include/dlt/dlt_common.h
razr: as far as I understood the source code, WIN32 ifdefs in those places was an attempt to commented out serial support which is not available on Windows.
I did it in a generic way by introducing DLT_SERIAL instead

razr:
2. I have replied to @ManikandanChockalingam comment.
I want to introduce DLT_SERIAL as an optional feature, at the moment it is hard-coded.

clipka:
@gunnarx  @razr 
I would prefer to have the change from include linux/limits.h to limits.h in a separate PR. This can be merged straight away from my point of view.

Regarding the DLT_SERIAL change, I can imagine that it makes sense to disable it at all, but then all test applications where your can specify the serial line, e.g. dlt-receive, should either have this option disabled as well or print a warning that serial is not configured.

gunnarx:
@razr can you split the PR as requested by clipka?

clipka:
Hi @razr, are you still interested in bringing the WITH_DLT_SERIAL option into the code base? I believe the other part about the includes cleanup is already part of the master branch.

ssugiura:
@razr This is a friendly reminder. If you are still interested in bringing WITH_DLT_SERIAL option, please rebase your code to our latest master. Otherwise, please close this PR.

ssugiura:

PR #43, opened 2018-02-21
IPC: Unix socket added

ManikandanChockalingam:
The user can select either FIFO or UNIX socket as
IPC between user library and daemon through
CMakelist option.

Signed-off-by: ManikandanC <Manikandan.Chockalingam@in.bosch.com>


src/daemon/dlt_daemon_common.c
```diff
@@ -295,13 +295,21 @@ int dlt_daemon_applications_clear(DltDaemon *daemon, int verbose)
     return 0;
 }
 
-DltDaemonApplication* dlt_daemon_application_add(DltDaemon *daemon, char *apid, pid_t pid, char *description, int verbose)
+DltDaemonApplication* dlt_daemon_application_add(DltDaemon *daemon,
+                                                 char *apid,
```
gunnarx: Please separate indentation/whitespace changes from code changes.
If you are going to reformat, why not reformat all code at once instead?
(synchronized with maintainer @clipka of course).

include/dlt/dlt_common.h
```diff
@@ -1084,6 +1084,13 @@ extern "C"
      * @return number of received bytes or negative value if there was an error
      */
     int dlt_receiver_receive_fd(DltReceiver *receiver);
+    /**
+     * Reiceve data from file/fifo/socket, calls corresponding function based on
```
gunnarx: s/Reiceve/Receive/

clipka:
@gunnarx Runtime decision seems bit difficult to us (even if preferred), because this decision then has to be made inside the dlt library at runtime as well. There might the case that an application is started before the DLT Daemon and then expects to write to FIFO, but the Daemon opens the UNIX socket. If you have an idea how to overcome the problem, we are happy to hear about.

gunnarx:
Not right away, but different choices seem to lead to different advantages and disadvantages.  I feel that once #ifdefs are introduced, they tend to get stuck there and it should be considered very carefully first.
With compile-time decision the library and daemon must be compiled identically, so that can also lead to issues similar to the ones you mention.   

Should integrators be able to choose the behavior not during compilation but during the system integration (e.g. environment variable?)

I also saw some of the thread that had a bit of discussion on this but wasn't sure of the main driver behind the UDS introduction.  Is it primarily being able to add more security than just protecting the FIFO file name?  If you intend to pass process credentials (UID/GID) then it makes sense.  UDS might be a a solution if the security is to be based on that.

A bunch of design decisions that might be worth discussing here.  Maybe mailing list is better?

clipka:
Your point of system integration time makes sense to me. I will start a discussion internally about your point. 
Another reason to have UNIX socket instead of FIFOs is the ability to run DLT applications inside a container.

CMakeLists.txt
```diff
@@ -247,6 +256,7 @@ message( STATUS "CMAKE_HOST_SYSTEM_PROCESSOR = ${CMAKE_HOST_SYSTEM_PROCESSOR}" )
 message( STATUS "CMAKE_SYSTEM_PROCESSOR = ${CMAKE_SYSTEM_PROCESSOR}" )
 message( STATUS "WITH_DLT_LOGSTORAGE_CTRL_UDEV = ${WITH_DLT_LOGSTORAGE_CTRL_UDEV}" )
 message( STATUS "WITH_DLT_LOGSTORAGE_CTRL_PROP = ${WITH_DLT_LOGSTORAGE_CTRL_PROP}" )
+message( STATUS "DLT_IPC = ${DLT_IPC} (Path: ${DLT_USER_SOCKET_PATH})" )
```
tobsan: `DLT_SOCKET_PATH` is would only be set if the IPC is `UNIX_SOCKET`

PR #44, opened 2018-03-12
Update dlt-system-syslog.c

sebc1686dev:
Initialize the variable t_attr before being used by pthread_create.

### DIFF
--- a/src/system/dlt-system-syslog.c
+++ b/src/system/dlt-system-syslog.c
@@ -165,6 +165,7 @@ void start_syslog(DltSystemConfiguration *conf)
 			DLT_STRING("dlt-system-syslog, start syslog"));
 	static pthread_attr_t t_attr;
 	static pthread_t pt;
+	pthread_attr_init(&t_attr);
 	pthread_create(&pt, &t_attr, (void *)syslog_thread, conf);
 	threads.threads[threads.count++] = pt;
 }


gunnarx:
PRs #44, #45, #46, #47, #48, #49 make the same single-line change to different

You should definitely send that as one single PR, not one for each.  In my personal opinion, all these changes are related and therefore should also be one single commit.  

I will comment on this one regarding the code and same comments apply for the other cases.

- "Update" is not a good first-line description on the commit.

src/system/dlt-system-syslog.c
```diff
@@ -165,6 +165,7 @@ void start_syslog(DltSystemConfiguration *conf)
 			DLT_STRING("dlt-system-syslog, start syslog"));
 	static pthread_attr_t t_attr;
 	static pthread_t pt;
+	pthread_attr_init(&t_attr);
```
gunnarx: Having t_attr uninitialized definitely seems like a bug so I agree with fixing it, but:

- pthread_create() man page says you can instead pass NULL to get the default attributes (which is what this also does, since t_attr is not modified after init call).

- pthread_attr_init() man page says you should destroy t_attr using pthread_attr_destroy() when it is no longer used, which is not done here.

For simplicity, I'd go with NULL.

PR #50, opened 2018-04-04
Update in dlt_daemon_client.c

kundatipradeep:
Even Forcing log level and trace status of context to not exceed "ContextLogLevel" and "ContextTraceStatus" in dlt.conf file with ForceContextLogLevelAndTraceStatus = 1, the default log level is being overwritten with high value than expected by client(DLT_Viewer when default trace data request is sent ) and by dlt-control application on usage of (dlt-control -d loglevel IP_ADDRESS)

With provided patch/changes in place, when ForceContextLogLevelAndTraceStatus  is enabled the requested log level is checked with ContextLogLevel and if it satisfied the daemon default loglevel gets updated.

### DIFF
--- a/src/daemon/dlt_daemon_client.c
+++ b/src/daemon/dlt_daemon_client.c
@@ -1747,7 +1747,10 @@ void dlt_daemon_control_set_default_log_level(int sock, DltDaemon *daemon, DltDa
     if (/*(req->log_level>=0) &&*/
             (req->log_level<=DLT_LOG_VERBOSE))
     {
-        daemon->default_log_level = req->log_level; /* No endianess conversion necessary */
+        if(daemon_local->flags.enforceContextLLAndTS)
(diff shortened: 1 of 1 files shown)


kundatipradeep:
Hi Mani,

I was about to ask the same question, reg "using set log level /dlt-control -l log level", once after getting the review comments from you.
I was in assumption in that, it was an requirement that even post setting "ForceContextLogLevelAndTraceStatus " in dlt.conf file users are given privilege to update the loglevel of registered context using dlt-control -l option. If its not intended I have a changes ready for individual log level and will update the changes accordingly along with considering review comments..

clipka:
Hi,

If it is forced by configuration, nobody should be able to override it from outside. This would not make sense. If this should be possible, the "forceLoglevel" can be disabled.

PR #58, opened 2018-05-11
dlt-daemon: Lower log level of logs not to output unintentional warning

ysk-sato:
Change log level of several logs from WARNING to INFO.
1. User FIFO cannot be opened because the FIFO does not exist
   --> This case can happen if DLT user app is one-shot
2. Runtime configuration file (dlt-runtime-application.cfg / dlt-runtime-context.cfg)
   cannot be opened
   --> This case can happen in normal use case.


ysk-sato:
@clipka I pushed new patch set. kindly please have look again.

PR #60, opened 2018-05-11
dlt-daemon: Output current number of connections as default

ysk-sato:
In previous, number of connections was output only if verbose mode
in dlt.conf was enabled. This information is basically useful
and is not output frequently. Therefore number of connections information
is made to output as default when new connection is established or
the connection is lost.

### DIFF
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -1483,6 +1483,7 @@ int dlt_daemon_process_client_connect(DltDaemon *daemon,
     struct sockaddr cli;
 
     int in_sock = -1;
+    char local_str[DLT_DAEMON_TEXTBUFSIZE] = { '\0' };
 
     PRINT_FUNCTION_VERBOSE(verbose);
 
@@ -1532,15 +1533,6 @@ int dlt_daemon_process_client_connect(DltDaemon *daemon,
         return -1;
     }
 
-    if (daemon_local->flags.vflag)
-    {
-        snprintf(str,
-                 DLT_DAEMON_TEXTBUFSIZE,
-                 "New connection to client established, #connections: %d\n",
-                 daemon_local->client_connections);
-        dlt_log(LOG_INFO, str);
-    }
-
     // send connection info about connected
     dlt_daemon_control_message_connection_info(in_sock,
                                                daemon,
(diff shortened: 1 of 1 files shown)


src/daemon/dlt-daemon.c
```diff
@@ -1569,6 +1561,14 @@ int dlt_daemon_process_client_connect(DltDaemon *daemon,
         }
     }
 
+    snprintf(local_str,
+             DLT_DAEMON_TEXTBUFSIZE,
+             "New client connection #%d established, Total Clients : %d\n",
+             in_sock,
+             daemon_local->client_connections);
+    dlt_log(LOG_DEBUG, local_str);
```
clipka: please use dlt_vlog instead of snprinft/dlt_log pair

src/daemon/dlt-daemon.c
```diff
@@ -3224,6 +3224,8 @@ int create_timer_fd(DltDaemonLocal *daemon_local,
 /* Close connection function */
 int dlt_daemon_close_socket(int sock, DltDaemon *daemon, DltDaemonLocal *daemon_local, int verbose)
 {
+    char local_str[DLT_DAEMON_TEXTBUFSIZE] = { '\0' };
```
clipka: same as above

src/daemon/dlt-daemon.c
ysk-sato: By this patch, I tried to output new log as both DLT message(with Apid: DLTD, Ctid: INTM) and internal log(by dlt_log()) with DEBUG log level. In this case, I cannot replace snprintf()/dlt_log() pair to dlt_vlog. How do you think? I thought these information should be visible as DLT message to know current status of dlt-daemon.

src/daemon/dlt-daemon.c
ysk-sato: same as above.

src/daemon/dlt-daemon.c
ysk-sato: Yes I saw the dlt_vlog/dlt_log implementation :-)
I guess you missed change of line 1570..
dlt_daemon_log_internal() can send DLT message from dlt-daemon but this function requires constructed string as the argument.

clipka: ah yes, this is bad and should be changed later. than it is ok from my side.

PR #63, opened 2018-05-22
dlt-daemon: Avoid to output duplicated application registration message

ysk-sato:
If application tries to send application registration request before
dlt-daemon startup, the request is sent twice as below:
1. Buffered request in user library buffer
2. New request when user app succeeds to attach to dlt-daemon

In previous, dlt-daemon had been output the registration messages
2 times even though these messages were come from same application.
This duplicated output is avoided by this change.

Warning message in case of dlt-runtime-context.cfg exists
is also avoided. This file is created when Control message: "Store Config"
is used. This is not irregular case.


ysk-sato:
@ManikandanChockalingam 
Thank you for your feedback.
The change is line 360 in dlt_daemon_common.c.
Actually stored ID information is registered as PID: 0.
You can find that in line 572 in dlt_daemon_common.c.

ManikandanChockalingam:
Ah..fine. i have missed to notice it.

PR #65, opened 2018-05-23
dlt-daemon: Fix no state transition to BUFFER state

ysk-sato:
When epoll detected error for TCP client connection, no state transition
had been happened. (Only the file descriptor had been closed)
To prevent this, dedicated close function is called according to connection type.


ysk-sato:
@ManikandanChockalingam To see epoll error for TCP connection, for example we can use below command on same system where dlt-daemon is running.  

`dlt-control -d 4 localhost`

It looks POLLHUP and POLLERR can happen normally if socket was somehow closed or crashed. (Please also refer to (http://man7.org/linux/man-pages/man7/socket.7.html))

Kindly please also have a look my new patch which correct typo :-)

PR #74, opened 2018-08-06
Fix gcc 8 build

jeez:
This PR is providing 3 fixes that allow us to build dlt-daemon with GCC 8 when the stringop-truncation warning is treated as an error.

### DIFF
--- a/src/console/dlt-control-common.c
+++ b/src/console/dlt-control-common.c
@@ -210,7 +210,7 @@ int dlt_parse_config_param(char *config_id, char **config_data)
                         {
                             *(config_data) = (char*)
                                     calloc(DLT_DAEMON_FLAG_MAX, sizeof(char));
-                            strncpy(*config_data,
+                            memcpy(*config_data,
                                 value,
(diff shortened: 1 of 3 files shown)


jeez:
I believe this fixes https://github.com/GENIVI/dlt-daemon/issues/64

clipka:
Hi, 
thanks @jeez for your contribution. I can see other warnings like -Wsizeof-pointer-memaccess, -Wformat-truncation=, ... Wouldn't it make sense to fix all warnings produced by gcc-8 within this PR?

clipka:
Hi, 
thanks @jeez for your contribution. I can see other warnings like -Wsizeof-pointer-memaccess, -Wformat-truncation=, ... Wouldn't it make sense to fix all warnings produced by gcc-8 within this PR?

jeez:
@clipka Yes, there are some warnings left. My intention with this PR however was to fix the build so the project can be used on gcc-8 based distros again. The warning can be fixed on demand, specially because some of them require a bit more understanding of the original than what I currently have.

PR #75, opened 2018-08-10
build: Remove unused include that was breaking build

edersondisouza:
glibc 2.28 install new headers, accessible via `sys/stat.h` that
redefine things defined on `linux/stat.h`. As dlt-daemon doesn't appear
to use anything from `linux/stat.h`, just remove it.

Signed-off-by: Ederson de Souza <ederson.desouza@intel.com>


edersondisouza:
Yep, that should solve the issue. Thanks!

PR #77, opened 2018-08-25
Fix compilation with glibc 2.28

sashko:
Package compilation fails with GLIBC 2.28:

stat.h:56:8: error: redefinition of 'struct statx_timestamp'
stat.h:99:8: error: redefinition of 'struct statx'

Signed-off-by: Oleksandr Kravchuk <oleksandr.kravchuk@pelagicore.com>


gunnarx:
@clipka, what's up - did you take a look?  Needed for upcoming updates to baseline, etc. and we'd rather not keep a patch around unnecessarily.

PR #84, opened 2018-11-15
Gateway Improvements

ManikandanChockalingam:
-Support to send and parse periodic control messages
 -add application/contexts to passive ECU list
 -Refactor dlt_gateway_send_control_message
 -Gateway issues with corrupted data and on demand connection
 -DLT Service ID Enum instead of defines
 -dlt-control: update get log info
 -dlt-client: logging: Extended the receiver buffer size
 -Unit Test update

Signed-off-by: Saya Sugiura <ssugiura@jp.adit-jv.com>
Signed-off-by: S. Hameed <shameed@jp.adit-jv.com>
Signed-off-by: ManikandanC <Manikandan.Chockalingam@in.bosch.com>

### DIFF
--- a/include/dlt/dlt_client.h
+++ b/include/dlt/dlt_client.h
@@ -296,9 +296,9 @@ int dlt_client_set_socket_path(DltClient *client, char *socket_path);
  * Parse GET_LOG_INFO response text
  * @param resp      GET_LOG_INFO response
  * @param resp_text response text represented by ASCII
- * @return 0 on success, -1 otherwise
+ * @return Value from DltReturnValue enum
  */
-int dlt_client_parse_get_log_info_resp_text(DltServiceGetLogInfoResponse *resp,
+DltReturnValue dlt_client_parse_get_log_info_resp_text(DltServiceGetLogInfoResponse *resp,
                                             char *resp_text);
 
 /**
--- a/include/dlt/dlt_client.h
+++ b/include/dlt/dlt_client.h
@@ -180,6 +180,18 @@ DltReturnValue dlt_client_send_log_level(DltClient *client, char *apid, char *ct
  */
 int dlt_client_get_log_info(DltClient *client);
 /**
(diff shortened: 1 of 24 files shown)


gunnarx:
I see that developers, and all the maintainers have signed off on this already, and it's your project so you decide, ... but isn't that a bit much to put into one big commit?  

All the listed things seem to me like they are independent, and should therefore be individual commits.  Otherwise it is just difficult to take them apart if there is some trouble later on and also it's not that convenient for people to review it all in one go either.

ManikandanChockalingam:
@gunnarx  i have made the changes to multiple commits. Please check and let me know if its fine with you

clipka:
@gunnarx, you are in principle right. We should make sure that we update the OSS code more frequently to have smaller commits and make out lives easier. But this did not happened during last months. Furthermore from the experience, we do not really get comments on the code when pushing code for additional features like logstorage or multiNode. Therefore, I would not spend too much effort in splitting up those things more - just because there is not much benefit. 
The split @ManikandanChockalingam did was very easy to do, the rest is bit more complex.

gunnarx:
> ...make sure that we update the OSS code more frequently to have smaller commits

... this comment makes me confused about what development method you have internally - and I don't think these principles have anything to do with OSS license or not?

> get comments on the code when pushing code for additional features like logstorage or multiNode

That's an occasion where it is extra important to do this since it sounds like it would have quite a large impact on the code.  For sure, it is frustrating to not get more feedback from users, but it must also be easy for them to review.  I'm thinking also that DLT used in production and semi-production environments and someone else than the core team might like to roll back only some changes but keep some bugfixes, etc.

> But this did not happened during last months. 

I understand. If "the damage is done" already and it is complicated to split, then you are right to not spend time on it that can be spent better on other things, and just to think about it in the future.  Again, it's your call as maintainer(s) and it sounds like you should merge this then and we move on.

Issue #85, opened 2018-11-16
Dlt-logstorage multiple file creation error.


QQruZZnick:
**Environment:**
dlt daemon 2.17

**dlt.conf:**
OfflineLogstorageMaxDevices = 3
OfflineLogstorageDirPath = .
OfflineLogstorageMaxCounter = 100
OfflineLogstorageCacheSize = 500000

**dlt_logstorage.conf:**
[FILTER1]
LogAppName=LOG, LOG1
ContextName=.*
LogLevel=DLT_LOG_INFO
File=testlog
FileSize=99
NOFiles=5

**Steps:**
1. Build daemon and put dlt.conf and dlt_logstorage.conf to dlt-daemon executable directory
2. Build 2 copies of dlt-example-user with different app-id's "LOG" and "LOG1"
3. Start dlt-daemon
4. Start dlt-example-user1 -n 1000 -d 500 testlog1
5. Start dlt-example-user2 -n 1000 -d 1000 testlog2

**Expected:**
Dlt-daemon creates 5 files (as specified in dlt_logstorage.conf) and rotates them.

**Error:**
Dlt-daemon creates 10 files and rotates them, 5 files for "LOG" app id and 5 files for "LOG"1 app id.

**Reason:**
Bad synchronization of dlt_offline_logstoarge files.
We need to look at struct **DltLogStorage** inside the **DltDaemon** struct. It has **config_list** which is key-value list. Each value **DltLogStorageFilterConfig** has field **records**. It is the list of files that was recorded in offline logstorage directory, and it used for creating new file when the old one is full. Than we look to the **dlt_logstorage_open_log_file** function:
```
if (config->records == NULL)
{
    if (dlt_logstorage_storage_dir_info(file_config, storage_path, config) != 0)
    {
        return -1;
    }
}
```
This code executes when log with key (CTXT: or APPID: or APPID:CTXTID) writes in logstorage for the first time. **dlt_logstorage_storage_dir_info** fills **records** with names of log files which are present in directory at this moment. But another key has it's own **records** and it can rotate files in directory which wont be actual for other keys. So dlt-daemon rotates files for each key but not for each [FILTER]
The fastest fix is:
```
erase(config->records)
config->records = NULL;
if (config->records == NULL)
{
    if (dlt_logstorage_storage_dir_info(file_config, storage_path, config) != 0)
    {
        return -1;
    }
}
```
But it will do a lot of unnecessary actions and will load CPU.

[...]

PR #87, opened 2018-11-23
Logstorage bug fix

ManikandanChockalingam:
fix for #85


ManikandanChockalingam:
There is problem with the fix. we are working on it internally.

PR #95, opened 2019-01-17
libdlt: Do not allow DLT usage in forked child

sherkenhoff:
DLT shall not be used in a forked child until a variant of exec() is
called, because DLT is using non async-signal-safe functions.

The forking process can continue to use libdlt's logging facilities, but
any attempt to use libdlt from the forked child will be denied.

The fork-handler test is updated to reflect this by trying to log from
the forked child which will fail. The fork then calls exec and runs
another application that can continue to use DLT.

Signed-off-by: Simon Herkenhoff <sherkenhoff@jp.adit-jv.com>


sherkenhoff:
@clipka @ManikandanChockalingam please review and merge.

PR #103, opened 2019-02-05
Fix linking problem with tests when systemd enabled

jardous:
* fixes sd_booted linking problem
* fixes systemd linking problem


jardous:
@clipka could you please review and merge this? The previous linking issue fix just introduced the systemd linking problem. This PS should solve all.

Would you prefer, if I create an Issue for the systemd linking problem?

PR #104, opened 2019-02-14
DLT_FORK_Child_LOCKUP

kundatipradeep:
As current implementation takes care of not Initializing the new set of buffers for the forked process, but the forked process would still contains the copy of shared resources and memory, which were not freed during the exit of child process. Calling dlt_free from child process would result in dead lock situation if parent process has locked on the resource during the child creation. So, to overcome the dead lock situation, allocating the memory for mutex at run time instead being the static and freeing it from the child process and deleting it only from the parent process and added a new function dlt_force_free to free the resources from the child process which doesn't have any locking mechanism.

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -89,7 +89,7 @@ static char dlt_daemon_fifo[NAME_MAX + 1];
 
 static char str[DLT_USER_BUFFER_LENGTH];
 
-static sem_t dlt_mutex;
+static sem_t *dlt_mutex;
 static pthread_t dlt_receiverthread_handle;
 
 /* calling dlt_user_atexit_handler() second time fails with error message */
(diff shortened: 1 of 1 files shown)


sherkenhoff:
@kundatipradeep The purpose of fork and exec is to cleanly start a child process. Posix discourages the use of any async unsafe calls in the fork before calling exec. Once exec is called, any memory present in fork is released and a new process is created. Could you please clarify your use case?

kundatipradeep:
@sherkenhoff I was not complete aware that exec followed by fork would release the memory in fork rather I was in assumption that the memory would get overwritten by the program loaded.

And apart from it I believe that fork and exec not required to be used together. It's should be acceptable for a program to fork itself without 'exec' ing, in which scenario, the child fork handler at current implementation doesn't the free the memory, and other file descriptors for the child process, which results in memory leak of child process. Calling dlt_free() in current scenario hangs due to lock mechanism. Please let me know if you have any queries further..

sherkenhoff:
@kundatipradeep 

> lease let me know if you have any queries further..

I'm interested in the use case. What are you going to do in the forked child apart from calling exec? 

> And apart from it I believe that fork and exec not required to be used together

Have you read about fork and exec, e.g. [this](https://www.evanjones.ca/fork-is-dangerous.html)?

ssugiura:
@kundatipradeep 
Are you still interested on this? If so, please share us your use case which requires this modification.

ssugiura:
@kundatipradeep This is a friendly reminder. Please resolve the conflicts and fix commit messages (please refer to discussion in https://github.com/GENIVI/dlt-daemon/pull/123 for detail) if you are still interested in bringing the feature.

ssugiura:

PR #105, opened 2019-02-14
DLT_FORK_CHILD_LOCKUP

kundatipradeep:
Added a new prototype dlt_force_free which doesn't have locking mechanism, and dlt_mutex to a pointer instead being the static varibale


ssugiura:

PR #110, opened 2019-03-05
Pipe name to use APID instead of PID

jardous:
Currently DLT library uses PID of the process as a part
of the pipe's name which is created in dltpipes folder.

This should be changed to support PID namespace which means
that PID of the process can't be part of the name. We propose to
use DLT APP ID as a replacement because that is always unique per
user (client).

### DIFF
--- a/include/dlt/dlt_user.h
+++ b/include/dlt/dlt_user.h
@@ -486,6 +486,8 @@ DltReturnValue dlt_user_trace_network_segmented(DltContext *handle,
  */
 DltReturnValue dlt_init();
 
+DltReturnValue dlt_init_with_appid();
+
 /**
  * Initialize the user lib writing only to file.
  * This function has to be called first, before using any DLT user lib functions.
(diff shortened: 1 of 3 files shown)


ssugiura:
Hello @jardous ,
Can I get more background why you need this change? It depends what you want to achieve, but you also have an option to use unix socket.
And, please update the documents as well.

jardous:
Hi @ssugiura!

The reason for this is that in our setup, some of the clients are ran in a container. So we face the problem that it is not only hard to identify the client but there is a danger, that those processes have the same PID.

The documentation update will follow.

clipka:
@jardous because of the container use case, we have added the option to use unix sockets instead of pipe. Have you looked into this possibilitiy?

jardous:
@clipka Yes, we have also been thinking about this possibility. That would need some more time to prove if that is going to work for us though...

clipka:
@jardous, I am still not completely convinced. How can you be sure that all APIDs are unique in all started containers?

jardous:
@clipka well, that is up to the users. It is easy to change APID if there is a conflict. This is not the case of PID.

Also we have list of clients and they get their APID assigned.

BTW we did try the unix socket approach and there seems to be some issues. Will continue investigating.

Issue #90, opened 2018-12-03
POSIX shared memory APIs


karthikshanmugam:
The DLT implementation uses non-POSIX shared memory and semaphore API like shmget, semget etc.,
These APIs are not availble in POSIX, the suggestion is to use shm_open, sem_open APIs. I can submit a patch if needed.

visglz:
Is the patch already available somewhere, maybe as PR? I would also be interested in compiling DLT for non linux POSIX platforms.

clipka:
No, we do not have it in place for shared memory. You can compile it with communication over sockets for QNX by using current master.

ssugiura:
@karthikshanmugam Sorry that we haven't updated for long time. I want to make sure your expectation here. We do have an option for shared memory as IPC, but actually it depends on FIFO or unix socket to send the message header. So even shared memory is enabled, FIFO or unix socket is used, which won't solve the situation you mentioned https://github.com/GENIVI/dlt-daemon/issues/52. We do so because dlt-daemon needs to wait for both library and client, which for client only socket connection is available.
So the question is, is it OK to use FIFO/unix socket+shared memory, or you want the complete shared memory IPC?

karthikshanmugam:
Hello,

Sorry for the delayed response.

Having the header exchanged via socket will still not solve the problem.

Would POSIX message queue be an option instead of sockets to exchange
header and ShMem for data?

Best regards,
Karthik

ssugiura:
@karthikshanmugam I guess it's the same for message queue. We need to create a queue beforehand on QNX, which I think is no difference to the network stack problem.
So the option we have is, either getting another thread in dlt-daemon to wait on library message, or retrying socket creation in library (easier, but should not be part of upstreamed code in my feeling).

karthikshanmugam:
It is not same, the problem with the socket is that we cannot have two
network interfaces for an application. Message queues does not have this
limitation. It is okay to transfer header over message queue (instead of
sockets) and data over shared memory

ssugiura:
@karthikshanmugam Could you give me more details why you cannot have two network interfaces for an application? Also, I would like to know how you start network stacks mentioned in https://github.com/GENIVI/dlt-daemon/issues/52.
> Default Network Stack -> UDS communication between processess, launched earlier in the boot.
Debug Network Stack -> USB to Ethernet communication with debug IP address, launched later in the boot.
[...]

PR #113, opened 2019-03-27
using POSIX shared memory APIs (#90)

votrungchi:
Replaced all Linux specific shared memory APIs with POSIX alternatives.

Signed-off-by: Vo Trung Chi <Chi.VoTrung@vn.bosch.com>


ssugiura:
We will rework on this to use only shared memory.

PR #114, opened 2019-04-02
Active reattaching to dlt-daemon at start up

jardous:
Currently buffered messages in libdlt are not output,
if app outputs the log only before start up of dlt-daemon.
To resolve this, retrying to reattach to dlt-daemon is added
at start up of "dlt_segmented" thread.

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -2917,9 +2917,20 @@ void dlt_user_trace_network_segmented_thread(void *unused)
     prctl(PR_SET_NAME, "dlt_segmented", 0, 0, 0);
 #endif
 
+    int count = 0;
     s_segmented_data *data;
 
     while (1) {
+        // Normally reattaching to dlt-daemon is triggered when app tries to output log.
+        // Even if the app outputs log only before starting up of the dlt-daemon,
+        // this retry loop helps reattaching.
+        while ((dlt_user.dlt_log_handle < 0) && (count < DLT_USER_REATTACH_MAX_COUNT))
+        {
+            dlt_user_log_reattach_to_daemon();
+            usleep(DLT_USER_REATTACH_INTERVAL);
+            count++;
+        }
+
         /* Wait until message queue is initialized */
         dlt_lock_mutex(&mq_mutex);
 
--- a/src/lib/dlt_user_cfg.h
+++ b/src/lib/dlt_user_cfg.h
@@ -143,6 +143,13 @@
 /* Retry interval for mq error in usec */
 #define DLT_USER_MQ_ERROR_RETRY_INTERVAL 100000
 
+/* Maximum retry count for reattach to dlt-daemon
+   (Default: 30 count  * 500000 usec = 15s) */
+#define DLT_USER_REATTACH_MAX_COUNT 30
+
+/* Retry interval for reattach to dlt-daemon (1000 usec = 1ms) */
+#define DLT_USER_REATTACH_INTERVAL 500000
+
 
 /* Name of environment variable to change the dlt log message buffer size */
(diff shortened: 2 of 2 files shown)


src/lib/dlt_user.c
```diff
@@ -2917,9 +2917,20 @@ void dlt_user_trace_network_segmented_thread(void *unused)
...
 #endif
 
+    int count = 0;
     s_segmented_data *data;
 
     while (1) {
+        // Normally reattaching to dlt-daemon is triggered when app tries to output log.
+        // Even if the app outputs log only before starting up of the dlt-daemon,
+        // this retry loop helps reattaching.
+        while ((dlt_user.dlt_log_handle < 0) && (count < DLT_USER_REATTACH_MAX_COUNT))
```
votrungchi: This condition already checked in dlt_user_log_reattach_to_daemon(). 
As my opinion, you do not need to retry loop here, just do the same as dlt_user_receiverthread_function. That mean just call the dlt_user_log_reattach_to_daemon() before dlt_user_trace_network_segmented_thread_segmenter(data).

src/lib/dlt_user.c
```diff
@@ -2917,9 +2917,20 @@ void dlt_user_trace_network_segmented_thread(void *unused)
...
+        {
+            dlt_user_log_reattach_to_daemon();
```
ssugiura: I would propose to handle it somewhere else like in dlt_receiver thread because this situation is not only for network trace but also normal logging. The thing is we need to flush all the data in internal buffer to daemon after daemon started which is only done when new log message comes from application. Let us work on it a bit.

src/lib/dlt_user.c
```diff
@@ -2917,9 +2917,20 @@ void dlt_user_trace_network_segmented_thread(void *unused)
...
+            usleep(DLT_USER_REATTACH_INTERVAL);
```
ssugiura: As man page of usleep says we should use nanosleep instread, please do so from now on.

jardous:
@ysk-sato could you please share your comments about this PS?

src/lib/dlt_user.c
ysk-sato: As ssugiura also mentioned, we also need to flush normal log here.
If application does not send any network traces, network segmented thread is blocked on pthread_cond_wait() in line 2938. Therefore reattaching to daemon is tried before blocking point.

src/lib/dlt_user.c
ysk-sato: Yes, I also dlt_receiver thread can also do same thing. However I wanted to avoid adding new route for output log. If we do same thing from dlt_receiver thread, there will be 3 routes.
1. Output from Main thread (Thread of application)
2. Output from Network segmented thread
3. Output from Receiver thread
I thought this confuses us and it may have risk.

src/lib/dlt_user.c
ysk-sato: @jardous kindly please modify the code to use nanosleep if possible.

src/lib/dlt_user.c
jardous: @ysk-sato can you, please, provide the code for that?

ysk-sato: I meant that current source code already realized that. (= Already segmented thread will flush normal log). I thought that still we need retry loop here and we don't have to change code here.

ssugiura:
@jardous @ysk-sato Sorry to keep you waiting for long time. We had come up with solution that we change the functionality of "receiver" thread to "housekeeper". Please refer to https://github.com/GENIVI/dlt-daemon/pull/171.

jardous:
@ssugiura I am fine with a cleaner solution for that problem. Should I we close this pull request then?

ssugiura:
@jardous Thanks for your quick check. If you agree, please close this PR.

PR #115, opened 2019-04-04
Fixed memory leak when receiving network traces of 0xFFFF length

jardous:
0xFFFF was used as a magic number in the length field of a network
message in order to trigger a resend. DLT library will now use the
nw_trace_type field and a custom type to do this, thus assuring no
external client of the library can reach this
value(DLT_NW_TRACE_USER_DEFINED7).

### DIFF
--- a/include/dlt/dlt_types.h
+++ b/include/dlt/dlt_types.h
@@ -158,7 +158,7 @@ typedef enum
     DLT_NW_TRACE_USER_DEFINED4 = 0x0C,
     DLT_NW_TRACE_USER_DEFINED5 = 0x0D,
     DLT_NW_TRACE_USER_DEFINED6 = 0x0E,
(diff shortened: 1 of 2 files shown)


jardous:
Hello @clipka, could you, please, review this?

clipka:
hello @jardous . I am not working on DLT anymore. @ssugiura, could you please check? Thanks.

jardous:
Hi @ssugiura. I have pushed some new PS - do they fix that like what you ment?

ssugiura:
Hi @jardous I would prefer to change the macro DLT_NW_TRACE_USER_DEFINED7 to DLT_NW_TRACE_RESEND instead of introducing a new one, to follow AUTOSAR spec.

jardous:
@ssugiura fixed

PR #120, opened 2019-04-17
dlt-control: Wait until send buffer is empty before exiting

jardous:
By checking amount of unsent data in send buffer,
dlt-control will wait until send buffer is empty before exiting.
Thus, dlt-control ensure control message reaches to dlt-daemon.

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -2917,9 +2917,24 @@ void dlt_user_trace_network_segmented_thread(void *unused)
     prctl(PR_SET_NAME, "dlt_segmented", 0, 0, 0);
 #endif
 
+    int count = 0;
     s_segmented_data *data;
+    struct timespec reattach_interval;
+
+    reattach_interval.tv_sec = 0;
+    reattach_interval.tv_nsec = DLT_USER_REATTACH_INTERVAL;
 
     while (1) {
+        // Normally reattaching to dlt-daemon is triggered when app tries to output log.
+        // Even if the app outputs log only before starting up of the dlt-daemon,
+        // this retry loop helps reattaching.
+        while ((dlt_user.dlt_log_handle < 0) && (count < DLT_USER_REATTACH_MAX_COUNT))
+        {
+            dlt_user_log_reattach_to_daemon();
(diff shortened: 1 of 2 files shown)


ysk-sato:
Regarding this commit, I guess there was my name with "Signed-off-by:" statement in bottom of commit message. Could you add it?

jardous:
@ysk-sato yes - you are right, Yosuke! I am very sorry about that. It got stripped by accident. I will ad it immediately.

ssugiura:
I'm interested to understand the use case to reach this.

jardous:
This PS has been backported from an internal branch. It should solve problems with lost traces during a reset.

src/console/dlt-control.c
```diff
@@ -111,6 +112,22 @@ typedef struct {
...
+    /* Wait until send buffer is empty */
+    while (1)
+    {
+        int unsent;
+        ioctl(fd, TIOCOUTQ, &unsent);
+        if (!unsent)
+        {
+            break;
+        }
+        usleep(100000);
```
ssugiura: Timeout should be considered. There could be a situation where the user set timeout with -t option but enter this condition. It should be either put together with timeout or introduce a new option which should be isolated  (or ignore) from -t option.

ysk-sato: @jardous I guess you also have the another patch for adding timeout in your internal branch. It may be able to the answer to ssugiura's comment.

jardous: @ysk-sato I could find find any PS that adds related timeout to the dlt-control. Could you, please point me, which PS did you mean?

ysk-sato: The title would be "dlt-control: Fix infinite loop issue at waiting send buffer empty". Could you find again?

ssugiura:
Is this a correct commit? Summary and commit message/changed files won't match. I guess you wanted to modify dlt-control.

ssugiura:
@jardous @ysk-sato As commented, could you please check whether this is a correct commit for this pull request?

jardous:
@ssugiura yes - you are right, this commit was messed up with a different one. Will delete this and send new pull request. Sorry for that

PR #121, opened 2019-04-24
Unlock buffer on termination

votrungchi:
Sometimes it happens that internal threads present
in DLT library are cancelled with one of the task
still holding the mutex, then the main task gets locked forever.

The following changes have been made:

  - Cleanup on thread cancellation with pthread_cleanup_push() is
    introduced

  - Locking on the message queue lock and subsequent signaling in
    dlt_stop_threads is only done if the segmented thread was created.

  - It is unnecessary to have a robust mutex for the message queue.
    Replace it with a normal mutex and remove EOWNERDEAD logic. Add a
    pthread_cancel handler to the segmented thread that might hold the
    message queue mutex.

  - Message queue mutex is statically initialized

Signed-off-by: S. Hameed <shameed@jp.adit-jv.com>
Signed-off-by: Jeevan Ramakant Nagvekar <jeevan.nagvekar1@wipro.com>
Signed-off-by: Saya Sugiura <ssugiura@jp.adit-jv.com>

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -106,19 +106,17 @@ static int g_dlt_is_child = 0;
 #define DLT_MESSAGE_QUEUE_NAME "/dlt_message_queue"
 #define DLT_DELAYED_RESEND_INDICATOR_PATTERN 0xFFFF
 
+#define DLT_UNUSED(x) (void)(x)
+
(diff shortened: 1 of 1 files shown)


ssugiura:
@votrungchi Please modify the commit message as suggested by @sherkenhoff .

PR #123, opened 2019-05-14
Made writev reliable for fixing data corruption

KalyaniPeddireddygari:
On random scenario's  "writev()" writes only part of the DLT data to Daemon FIFO due to un-availability of space in Daemon FIFO,  and returns PIPE Error which in turn stores the same complete trace in the applications ring buffer. 
Incomplete trace was read by Daemon(from FIFO) and routed the same to DLT Clients which in turn results in Data corruption at DLT Viewer due to non-availability of complete trace as per the DLT Payload headers.  
Hence proposed solution would check for space availability for each trace prior writing the data to Daemon FIFO, during the failure conditions the data to be stored internally in its application ring buffers and ensuring incomplete traces were not written on to Daemon FIFO.

### DIFF
--- a/src/shared/dlt_user_shared.c
+++ b/src/shared/dlt_user_shared.c
@@ -73,9 +73,66 @@
 #include <errno.h>
 
 #include <sys/uio.h> /* writev() */
+#include <sys/ioctl.h> /* ioctl() */
+#include <string.h> /* strerror() */
+#include <syslog.h>
 
 #include "dlt_user_shared.h"
 #include "dlt_user_shared_cfg.h"
+
+static DltReturnValue dlt_writev_reliable(int handle, void *ptr1, size_t len1, void *ptr2, size_t len2, void *ptr3, size_t len3);
+
+DltReturnValue dlt_writev_reliable(int handle, void *ptr1, size_t len1, void *ptr2, size_t len2, void *ptr3, size_t len3)
+{
+    struct iovec iov[3];
+    uint32_t bytes_written;
+    int daemon_fifo_size;
+    size_t non_read_bytes = 0;
+
+    if (handle <= 0)
+        /* Invalid handle */
+        return DLT_RETURN_ERROR;
+
+    iov[0].iov_base = ptr1;
+    iov[0].iov_len = len1;
+    iov[1].iov_base = ptr2;
+    iov[1].iov_len = len2;
+    iov[2].iov_base = ptr3;
+    iov[2].iov_len = len3;
+
+    if (dlt_is_file_check != 1)
+    {
+        /* Get Daemon FIFO size */
+        if ((daemon_fifo_size = fcntl(handle, F_GETPIPE_SZ , 0)) == -1)
+        {
+            dlt_vlog(LOG_ERR, "get FIFO size error, could not perform write operation: %s\n",strerror(errno));
+            return DLT_RETURN_ERROR;
+        }
+        if ((daemon_fifo_size > 0) && (ioctl(handle, FIONREAD, &non_read_bytes) != -1))
+        {
+            // Determine the FIFO size availability prior writing data to PIPE
+            if ((daemon_fifo_size - non_read_bytes) > (len1+len2+len3))
+            {
+                bytes_written = writev(handle, iov, 3);
+            }
+            else
+            {
+                dlt_vlog(LOG_DEBUG, "Not enough space available to write the data to FIFO, hence storing it in ring buffer\n");
+                return DLT_RETURN_ERROR; // Not enough size to write data to FIFO
+            }
+        }
+        else
+            return DLT_RETURN_ERROR;
+    }
+    else
+    {
+        // writing data to a File
+        bytes_written = writev(handle, iov, 3);
+    }
+    if (bytes_written != (len1 + len2 + len3))
+        return DLT_RETURN_ERROR;
+    return DLT_RETURN_OK;
+}
 
 DltReturnValue dlt_user_set_userheader(DltUserHeader *userheader, uint32_t mtype)
 {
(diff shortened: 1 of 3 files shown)


src/shared/dlt_user_shared.c
```diff
@@ -143,7 +149,30 @@ DltReturnValue dlt_user_log_out3(int handle, void *ptr1, size_t len1, void *ptr2
...
+        snprintf(str,sizeof(str),"get FIFO size error, could not perform write operation: %s\n",strerror(errno));
+        dlt_log(LOG_ERR, str);
+        return DLT_RETURN_ERROR;
+    }
+    if (daemon_fifo_size > 0)
+    {
+        if (ioctl(handle, FIONREAD, &non_read_bytes) != -1)
+        {
+            // Determine the FIFO size availability prior writing data to PIPE
+            if ((daemon_fifo_size - non_read_bytes) > (len1+len2+len3))
```
votrungchi: Use FIONSPACE to get the FIFO free space.

src/shared/dlt_user_shared.c
```diff
@@ -143,7 +149,30 @@ DltReturnValue dlt_user_log_out3(int handle, void *ptr1, size_t len1, void *ptr2
...
+            {
+                bytes_written = writev(handle, iov, 3);
+            }
+            else
+            {
+                snprintf(str,sizeof(str),"Not enough space available to write the data to daemon FIFO, hence storing it in ring buffer\n");
+                dlt_log(LOG_DEBUG, str);
```
votrungchi: Does not need the str here, you can log the message string directly with dlt_log(). In additional, if you need logging with format string, please use dlt_vlog().

src/shared/dlt_user_shared.c
KalyaniPeddireddygari: @votrungchi , I believe FIONSPACE is specific to QNX OS, I don't think it can be used on Linux. Please correct me if I am wrong, and let me know if there is other alternative to be used.

src/shared/dlt_user_shared.c
KalyaniPeddireddygari: @votrungchi updated dlt_log with dlt_vlog as per suggestion in new commit

src/shared/dlt_user_shared.c
```diff
@@ -73,6 +73,9 @@
 #include <errno.h>
 
 #include <sys/uio.h> /* writev() */
+#include <sys/ioctl.h> /* ioctl() */
+#include <string.h>  /* strerror() */
+#include <syslog.h>
```
sherkenhoff: Remove trailing white space.

sherkenhoff:
Instead of testing how many bytes are available to write in the FIFO and return an error, dlt_user_log_out should note the number of bytes written and attempt to write the remaining bytes before returning. For posix, partial writes are not an error case and we should not look at it this way.

Also, the commit only touches dlt_user_log_out3, but we also have dlt_user_log_out2. I further propose to do the write / send remaining / error checking in one central function.

@ssugiura Any thoughts?

KalyaniPeddireddygari:
@sherkenhoff : Actually We thought about it(attempt to write the remaining bytes before returning) and had a simulated model at our end, where in the behavior was not consistent.
There were cases where the partial write had happened and in its sub-sequent write(writing remaining bytes) resulted in EPIPE or EAGAIN errors, which is not intended behavior. So we were returning the DLT_RETURN_ERROR when partial write had happened.

We shall consider having the writev under one central function for dlt_user_log_out2 and dlt_user_log_out3 and will submit the changes asap.

src/shared/dlt_user_shared.c
KalyaniPeddireddygari: Updated as per comments.

sherkenhoff:
@KalyaniPeddireddygari We try to keep commit messages similar to Linux Kernel commit messages (see e.g. [here](https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#the-canonical-patch-format)). Also, please add a signed-off by line and and squash / split your changes into logically consistent commits. I.e. there is no need to add a new commit to remove a trailing white space. This just unnecessarily pollutes the main branch.

KalyaniPeddireddygari:
@sherkenhoff :  Squashed all the changes into a single commit, Please provide your inputs on the same.

src/shared/dlt_user_shared.h
```diff
@@ -75,6 +75,9 @@
 
 #include <sys/types.h>
 
+// Global variable to check the dlt logging is to a file or to FIFO
```
gunnarx: First, if you must use globals, it should still be defined in a .c file so it isn't defined more than once (if the header is used more than once).  

Then, if it is necessary to exchange this information between dlt-user.c and dlt_user_shared.c it might be nicer to provide a function in dlt-user.c that will answer this question (a function that can be called from dlt_user_shared.c)
This is basic encapsulation, and among other things it would "write protect" the variable, I mean prevent other modules than dlt-user from modifying it.

gunnarx:
If you follow the rule in the "Linux" style that was requested (and most other open source projects including the [_recommendation_ for all GENIVI projects](https://at.projects.genivi.org/wiki/x/ZwHe) ), then instead of:

```
    Updated Made writev reliable for fixing data corruption

    On random scenario's  "writev()" writes only part of the DLT data to Daemon FIFO due to un-availability of space in Daemon FIFO,  and returns PIPE Error which in turn stores the same complete trace in the applications ring buffer.
    Incomplete trace was read by Daemon(from FIFO) and routed the same to DLT Clients which in turn results in Data corruption at DLT Viewer due to non-availability of complete trace as per the DLT Payload headers.
    Hence proposed solution would check for space availability for each trace prior writing the data to Daemon FIFO, during the failure conditions the data to be stored internally in its application ring buffers and ensuring incomplete traces were not written on to Daemon FIFO.
```
it would look like this:
```
    Updated Made writev reliable for fixing data corruption
    
    On random scenarios "writev()" writes only part of the DLT data to
    Daemon FIFO due to un-availability of space in Daemon FIFO, and returns
    PIPE Error which in turn stores the same complete trace in the
    applications ring buffer. Incomplete trace was read by Daemon(from FIFO)
    and routed the same to DLT Clients which in turn results in Data
    corruption at DLT Viewer due to non-availability of complete trace as
    per the DLT Payload headers. Hence proposed solution would check for
    space availability for each trace prior writing the data to Daemon FIFO,
    during the failure conditions the data to be stored internally in
    its application ring buffers and ensuring incomplete traces were
    not written on to Daemon FIFO.
```

- You were also asked already to add a sign-off line, right?

Just to help, may I add the following minor reflections as well:

- You might notice that the word "Updated" is unnecessary noise here.  You are proposing a change that is not yet merged, so seen from the master branch... it "knows" nothing about this yet.  You can simply write as concisely and clearly as possible, what the change does.

- The consensus among users of the previous commit format is to use present (or imperative) tense, not past tense.  So: "Make writev reliable" instead of "Made writev reliable".  This is picky, but those are widely agreed upon "rules" so it might be good to get use to them.

sherkenhoff:
Thanks @gunnarx for taking so much time here :-)

@KalyaniPeddireddygari Thanks for your continuous contribution 👍 

Let's step back for a second and look at what your use case and problem is. To my understanding, you have a case where an application sends a message to the DLT daemon, but only a part of the data is transferred because the underlying buffer is full. This leads to some corruption in the viewer., because the viewer doesn't understand a partial message. In your opinion, the problem is due to the application side (libdlt) not checking the size available in the buffer, so you add a fcntl to check whether or not you can write a complete message into the buffer.

Let me try to argue for a different solution to this problem: Every message has a header that stores its length. We can modify the daemon to drop partial messages and let the application re-transmit the complete message at a later point (message kept in application side buffer until sent). Furthermore, I don't like the use of F_GETPIPE_SZ because there might be hundreds of applications writing to this single buffer. Imagine one application determines the buffer has enough free space, but then the scheduler runs a different thread that writes to the buffer. Your logic would fail. Also, the fcntl adds processing time to the application. A simple writev would suffice. I like the refactoring of writev into a single function, just maybe the name "dlt_writev_reliable" is not a good choice. How about "dlt_transfer_to_daemon", or simply "dlt_writev"?

Please, add a signed-off by line (git commit -s) to your commit(s) and modify the commit message according to @gunnarx comments. Those long lines look ugly in "git log". 

@ManikandanChockalingam Cc

gunnarx:
> Thanks @gunnarx for taking so much time here :-)

You are welcome. I sometimes feel bad that I am not out supporting each code project as much as I would like since I am also involved in various high-level domain and Multi-OS integration projects in GENIVI.   (And we have over 90 repositories in github/GENIVI to cover...)    
**Dlt-daemon** feels like a very healthy project so I should probably focus more on other projects. ;-)

ssugiura:
@KalyaniPeddireddygari First of all, we support both FIFO and unix socket, so please consider unix socket use case as well. But still, I think the situation is mostly the same.
In my feeling, it should be handled in daemon side, not library side as it will block the running application, because of logging.

@sherkenhoff @gunnarx Thanks a lot for your support.

sherkenhoff:
@KalyaniPeddireddygari Any updates here?

KalyaniPeddireddygari:
@sherkenhoff Couldn't spend time on this topic, will provide the feedback by next week.

ssugiura:
@KalyaniPeddireddygari Do you have any update on this?

ssugiura:
@KalyaniPeddireddygari Friendly reminder.

ssugiura:
@KalyaniPeddireddygari Are you still interested to bringing this fix in?

ssugiura:
Closing this PR since there's no update for long time. @KalyaniPeddireddygari Please reopen this PR if you're still willing to take the changes in. Thank you!

PR #128, opened 2019-05-25
adding dlt client example

yeger00:
I added a DLT client target, in order to be able and use it as command line utility.

The code is taken from the client tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ cd build
 cmake ..
 make
 optional: sudo make install
-optional: sudo ldconfig`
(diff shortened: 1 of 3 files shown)


ssugiura:
We already have dlt-receive which can do similar thing. Is there any difference?

aviyegerargus:
Looks the same. I'll add my changes to it, and close this PR.

Thanks.

yeger00:
Looks the same. I'll add my changes to it, and close this PR.

Thanks.

PR #129, opened 2019-05-27
Do not install systemd service files for binaries that are not built

martin-ejdestig:
Noticed that dlt-system.service unconditionally failed since the binary was missing when building with a default configuration after an upgrade from 2.17.0 to 2.18.1. WITH_DLT_SYSTEM default changed to OFF in 1b3c8b3d10fbc9a8dd9c717a939babd901b0b243 .

Fix by only installing dlt-system.service if dlt-system is built. Also do the same for other binaries that have a service file.

### DIFF
--- a/systemd/CMakeLists.txt
+++ b/systemd/CMakeLists.txt
@@ -30,31 +30,42 @@ if(WITH_SYSTEMD)
     configure_file(${CMAKE_SOURCE_DIR}/systemd/dlt.service.cmake ${PROJECT_BINARY_DIR}/systemd/dlt.service)
     message( STATUS "Configured systemd unit file:dlt.service" )
 
-    configure_file(${CMAKE_SOURCE_DIR}/systemd/dlt-system.service.cmake ${PROJECT_BINARY_DIR}/systemd/dlt-system.service)
-    message( STATUS "Configured systemd unit file:dlt-system.service" )
+    if(WITH_DLT_SYSTEM)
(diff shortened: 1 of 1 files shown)


systemd/CMakeLists.txt
```diff
@@ -30,31 +30,42 @@ if(WITH_SYSTEMD)
...
     endif(WITH_DLT_DBUS)
 
-    configure_file(${CMAKE_SOURCE_DIR}/systemd/dlt-receive.service.cmake ${PROJECT_BINARY_DIR}/systemd/dlt-receive.service)
-    message( STATUS "Configured systemd unit file:dlt-receive.service" )
+    if(WITH_DLT_CONSOLE)
+        configure_file(${CMAKE_SOURCE_DIR}/systemd/dlt-receive.service.cmake ${PROJECT_BINARY_DIR}/systemd/dlt-receive.service)
+        message( STATUS "Configured systemd unit file:dlt-receive.service" )
+    endif(WITH_DLT_CONSOLE)
 
     configure_file(${CMAKE_SOURCE_DIR}/systemd/dlt-example-user.service.cmake ${PROJECT_BINARY_DIR}/systemd/dlt-example-user.service)
```
ssugiura: Please add WITH_DLT_EXAMPLES condition as well.

martin-ejdestig: Done.

martin-ejdestig: Sorry, actually missed the "configure_file() if" clause and only did the "install() if" at first. Fixed now though.

Perhaps it makes sense to merge the if statements. Will try this in a separate commit.

ssugiura:
Thank you. I squashed all the commits and merged to master.

PR #130, opened 2019-05-30
Add option in dlt.conf for bindAddress to specific IPs

AdrianStoenescu:
dlt-daemon: Added functionality to use specific set of bindAddresses (IP addresses) 
for which the dlt-daemon will accept connections.

I added the possibility to have in dlt.conf an option that specifies a set of IP addresses.
The daemon will do a socket bind only for this list, therefore external connections will be 
limited to this set of IP addresses.
If this option is not given, the default INADDR_ANY is used ("0.0.0.0") as before.

The option in dlt.conf can look like this:
BindAddress = 160.48.199.98;160.48.199.97;160.48.199.226,160.48.199.186;160.48.199.139;172.16.222.99

There can be as many addresses but only specified on one single line. 
They have to be delimited by ',' or ';'. ' ' (space) delimiter does not work.
IPv6 addresses are supported as well. For instance:

BindAddress = fe80::255:7bff:feb5:7df7

There can be IPv4 addresses and IPv6 in the same time, but if at least one IPv6 address is used, 
then the dlt-daemon must be compiled with DLT_USE_IPv6 flag ON.

Signed-off-by: Adrian Stoenescu <Adrian.Stoenescu@harman.com>


ssugiura:
Summary and description lines are too long and there's no signed off. Please refer to https://github.com/GENIVI/dlt-daemon/pull/123 for commit message format.

src/daemon/dlt_daemon_socket.c
```diff
@@ -62,67 +63,64 @@
...
+    if (bind(*sock, (struct sockaddr *)&forced_addr, sizeof(forced_addr)) == -1) {
+        const int lastErrno = errno; /*close() may set errno too */
+        close(*sock);
+        dlt_vlog(LOG_WARNING, "dlt_daemon_socket_open: bind() error %d: %s\n", lastErrno, strerror(lastErrno));
+    }
 
-    snprintf(str, DLT_DAEMON_TEXTBUFSIZE, "%s: Listening on port: %u\n", __FUNCTION__, servPort);
+    /*listen */
+    snprintf(str, DLT_DAEMON_TEXTBUFSIZE, "%s: Listening on ip %s and port: %u\n", __FUNCTION__, ip, servPort);
     dlt_log(LOG_INFO, str);
```
ssugiura: Please use dlt_vlog instead of snprintf-dlt_log.

src/daemon/dlt_daemon_socket.c
```diff
@@ -62,67 +63,64 @@
...
+#endif
 
-        if (bind(*sock, p->ai_addr, p->ai_addrlen) == -1) {
-            const int lastErrno = errno; /*close() may set errno too */
-            close(*sock);
-            dlt_vlog(LOG_WARNING, "dlt_daemon_socket_open: bind() error %d: %s\n", lastErrno, strerror(lastErrno));
-            continue;
-        }
+    snprintf(str, DLT_DAEMON_TEXTBUFSIZE, "%s: Socket created\n", __FUNCTION__);
+    dlt_log(LOG_INFO, str);
```
ssugiura: Please use dlt_vlog instead of snprintf-dlt_log.

votrungchi:
Hi @AdrianStoenescu,

What happened if the "BindAddress" reached the limitation of socket of system?

As my opinion, to archive this purpose, we shouldn't bind all ip address in this list but We should filter the allowed ip address after accept the connection, if this client not in the list, we can close the accepted socket. (Check dlt_daemon_process_client_connect() function), Its also help to minimize the change.

Best regards,

AdrianStoenescu:
Hi @votrungchi 

Yes, your proposal seems good and logical. I will try to change the behavior as you mentioned. I will accept the connection and if the client address is not in my filter list, I will close it. In this way you don't do many bind calls that might not get a connection at all. Thank you!

AdrianStoenescu:
Hey @votrungchi and @ssugiura 
I thought a lot about having one bind to INADDR_ANY and after accept, decide if the client connected using an address from the filtered list specified in dlt.conf. It would work great, but after this PR I want to add the possibility to have specific ports as well associated with this bindAddresses. This means that I want each IP to have a specific Port, and connections are allowed only on these paris. In order to do this I will create socket/bind/listen for each pair IP/port. This means the current approach with binding for each IP is more reasonable. I don't think in practical cases somebody will use more then 2 or 3 IPs, so to have 2,3 bind calls is not much. What do you think?

src/daemon/dlt-daemon.c
```diff
@@ -566,6 +578,39 @@ int option_file_parser(DltDaemonLocal *daemon_local)
...
+                        DltBindAddress_t *head = daemon_local->flags.ipNodes;
+                        DltBindAddress_t *newNode = NULL;
+                        DltBindAddress_t *temp = NULL;
+
+                        /* update first IP node with address from configuration file */
+                        char *tok = strtok(value, ",;");
+
+                        if (tok != NULL) {
+                            /* set first IP address */
+                            strncpy(head->ip, tok, sizeof(head->ip) - 1);
```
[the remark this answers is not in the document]
AdrianStoenescu: If IPv6 is used, then a bind error will occur and dlt-daemon will not start.
The same for IPv4, if IPv4 is used with DLT_USE_IPv6 to ON, the same bind error will occur and dlt-daemon will not start

votrungchi:
Hi @AdrianStoenescu ,
First of all, the bindAddresses here are the address of the interface on the local machine (server) to listen on, it's not the client IP address. So I wonder if the implementation works as your expectation? If yes, could you please help to give me more detail about your expectation and how to test it?
Best Regrards,

AdrianStoenescu:
Hi @votrungchi 
Exactly, the limitation is on the server side. In other words, the dlt-daemon accepts connections ON the specified IP list. It has nothing to do with client address, which can be anything.
Example: 
dlt-daemon runs on a board with IP 192.168.1.100
BindAddress option in dlt.conf is:
BindAddress = 160.48.199.97;160.48.199.98

Attempt to connect with dlt-viewer using 192.168.1.100 will not work anymore, only connections using hostname 160.48.199.97 or 160.48.199.98.

If BindAddress option is not present in dlt.conf, the dlt-daemon will bind on INADDR_ANY, so will accept connections on all available interfaces. The dlt-viewer can connect using hostname 192.168.1.100, 160.48.199.97 160.48.199.98 or any available IP from a network interface that is up at that moment on the dlt-daemon side

votrungchi:
Hi @AdrianStoenescu ,
Yes, if I understand correct, you want to provide the feature to support Server to listen on specific addresses and ports, and I found the similar thing that supported in Apache HTTP Server with "Listen" configure keyword.
So I think you can refer to this for good practice in the implementation.
https://httpd.apache.org/docs/2.4/bind.html
Best Regards,

ssugiura:
@AdrianStoenescu Is there any update on this?

ssugiura:
@AdrianStoenescu FYI. I rebased to latest master and merged your features.

ssugiura:
As discussed with @AdrianStoenescu , I reverted this commit from master branch. We can push back to master once the feature is needed.

jardous:
Actually I would appreciate this to be merged

ssugiura:
@jardous Great to hear that. Is the current implementation already works for your use case? If so, I will take it back to master branch again.

ssugiura:
@AdrianStoenescu @jardous I merged this feature again to the master along with https://github.com/GENIVI/dlt-daemon/pull/152.

AdrianStoenescu:
OK, Mr. Sugiura

I will have a look and try to resolve the bug related IPv6 (when we are unable to bind using 0.0.0.0 address)

Adrian Stoenescu
Software Developer
Harman Connected Car Division

MetrOffice, Building A, 2nd District
5-7 Dimitrie Pompeiu Blvd.
020335 Bucharest
Phone: +40 372 518 259
Mobile: +40 726 734 437
Email: adrian.stoenescu@harman.com<mailto:adrian.stoenescu@harman.com>
Web: www.harman.com<http://www.harman.com/>
[cid:image001.jpg@01D2E525.19D2C600]

votrungchi:
> I will have a look and try to resolve the bug related IPv6 (when we are unable to bind using 0.0.0.0 address)
@AdrianStoenescu bind issue already fixed by #152 pull request.

PR #137, opened 2019-06-17
define DLT_PATH_MAX for max path buffer length

votrungchi:
define DLT_PATH_MAX for max path buffer length

DLT limits the path length and does not do anything else to determine
the actual value, because the least that is supported on any system
that DLT runs on is 1024 bytes.

Signed-off-by: Vo Trung Chi <Chi.VoTrung@vn.bosch.com>

### DIFF
--- a/include/dlt/dlt_common.h
+++ b/include/dlt/dlt_common.h
@@ -350,6 +350,14 @@ enum {
 #   define DLT_RCV_REMOVE      (1 << 1)
 
 /**
+ * Maximal length of path in DLT
+ * DLT limits the path length and does not do anything else to determine
+ * the actual value, because the least that is supported on any system
+ * that DLT runs on is 1024 bytes.
+ */
+#   define DLT_PATH_MAX 1024
+
+/**
  * Maximal length of mounted path
  */
 #   define DLT_MOUNT_PATH_MAX  1024
@@ -413,7 +421,7 @@ extern char dltSerialHeaderChar[DLT_ID_SIZE];
 /**
  * The common base-path of the dlt-daemon-fifo and application-generated fifos
  */
-extern char dltFifoBaseDir[PATH_MAX + 1];
+extern char dltFifoBaseDir[DLT_PATH_MAX];
 
 /**
  * The type of a DLT ID (context id, application id, etc.)
--- a/src/dbus/dlt-dbus-options.c
+++ b/src/dbus/dlt-dbus-options.c
@@ -82,21 +82,21 @@ int read_command_line(DltDBusCliOptions *options, int argc, char *argv[])
         {
             options->BusType = malloc(strlen(optarg) + 1);
             MALLOC_ASSERT(options->BusType);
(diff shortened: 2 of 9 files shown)


sherkenhoff:
> Since POSIX isn’t clear on path lengths, use PATH_MAX as a base for path + filename.

In POSIX, we can use [pathconf](http://pubs.opengroup.org/onlinepubs/009695399/functions/pathconf.html) with _PC_PATH_MAX - just verified on QNX (returns 1024 bytes) and on Linux (returns 4096 bytes). Is there any reason to not use this interface to determine the _proper_ value?

votrungchi:
Hi @sherkenhoff,

First, we need this value at the compile time, so if we can use the pathconf, it would be use in compile time.
I read about the PATH_MAX and pathconf in this [blog](https://eklitzke.org/path-max-is-tricky). Honestly, I not really sure about the idea of section "Path Metadata With pathconf()", what do you think?

Thanks & Best Regards,

sherkenhoff:
@votrungchi 

This is the same source I also read. Further down it explicitly states that the maximum length of a filename is eventually governed by the filesystem that is handling the node, and that PATH_MAX is only a lower bound, at best. With this in mind, I would not limit the length at compile time to the target's lower bound, but rather determine the actual value at run time at the very little cost of of the pathconf call. 

What would need to be changed in DLT to achieve determining the max path length at run time?

votrungchi:
Hi @sherkenhoff ,

Thank you for your detail explanation.
As my understand, we will use pathconf to determining the max path length at run time with path input is root directory "/". Is this correct?

For your question, to achieve determining the max path length at run time, will have to change all static memory allocation which size is PATH_MAX to dynamic memory allocation.

Best Regards,

sherkenhoff:
>we will use pathconf to determining the max path length at run time with path input is root directory "/"

No. "/" itself might also be mounted and handled by a filesystem that supports much longer names. If that were the case and somebody wanted to write the logs to a very nested location, he couldn't do it because DLT will now set the max length to 1024.

Adding to my question from before: How much load on the system would we introduce by changing to run-time determining the max length? I assume this happens once per process and once in the daemon?

include/dlt/dlt_common.h
```diff
@@ -377,6 +377,11 @@ enum {
  */
 #   define DLT_LINE_LEN 1024
 
+/**
+ * DLT fifo basedir length
+ */
+#define DLT_BASEDIR_MAX PATH_MAX - 16
```
ssugiura: Where did 16 come from? Either having another macro or give some explanation would be helpful.

src/dbus/dlt-dbus-options.c
```diff
@@ -30,6 +30,7 @@
 
 #include <stdlib.h>
 #include <string.h>
+#include <getopt.h>
```
thanhbnq: This header is not really needed here. Please don't add it.

votrungchi: Removed. Thanks.

PR #140, opened 2019-06-20
Fix clang warning

phongt:
mirror change for clang compiler warnings.
```console
$ mkdir build
$ cd build
$ cmake -DCMAKE_C_COMPILER=clang-8 -DCMAKE_CXX_COMPILER=clang++-8 -DCMAKE_BUILD_TYPE=Debug -DWITH_DLT_TESTS=ON -DWITH_TESTSCRIPTS=ON -DWITH_DLT_UNIT_TESTS=ON -DWITH_SYSTEMD=ON -DWITH_SYSTEMD_JOURNAL=ON ..
$ make
```
 gtest_dlt_* are fine.

### DIFF
--- a/src/console/dlt-convert.c
+++ b/src/console/dlt-convert.c
@@ -343,7 +343,10 @@ int main(int argc, char *argv[])
 
                         if (end == (file.counter - 1)) {
                             /* Sleep if no new message was received */
-                            sleep(0.1);
+                            struct timespec req;
+                            req.tv_sec = 0;
+                            req.tv_nsec = 100000000;
+                            nanosleep(&req, NULL);
                         }
                         else {
                             /* set new end of log file and continue reading */
--- a/src/shared/dlt_offline_trace.c
+++ b/src/shared/dlt_offline_trace.c
@@ -147,10 +147,10 @@ void dlt_offline_trace_file_name(char *log_file_name, char *name, unsigned int i
 
     /* create log file name */
     memset(log_file_name, 0, DLT_OFFLINETRACE_FILENAME_MAX_SIZE * sizeof(char));
-    strncat(log_file_name, name, sizeof(DLT_OFFLINETRACE_FILENAME_BASE));
-    strncat(log_file_name, DLT_OFFLINETRACE_FILENAME_DELI, sizeof(DLT_OFFLINETRACE_FILENAME_DELI));
-    strncat(log_file_name, file_index, sizeof(file_index));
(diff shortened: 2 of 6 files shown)


src/shared/dlt_offline_trace.c
```diff
@@ -149,7 +149,7 @@ void dlt_offline_trace_file_name(char *log_file_name, char *name, unsigned int i
     memset(log_file_name, 0, DLT_OFFLINETRACE_FILENAME_MAX_SIZE * sizeof(char));
     strncat(log_file_name, name, sizeof(DLT_OFFLINETRACE_FILENAME_BASE));
     strncat(log_file_name, DLT_OFFLINETRACE_FILENAME_DELI, sizeof(DLT_OFFLINETRACE_FILENAME_DELI));
```
votrungchi: This is also size of the source.

src/console/dlt-convert.c
```diff
@@ -343,7 +343,7 @@ int main(int argc, char *argv[])
 
                         if (end == (file.counter - 1)) {
                             /* Sleep if no new message was received */
-                            sleep(0.1);
+                            usleep(100000);
```
ssugiura: Please use nanosleep instead to follow current POSIX standard.

src/shared/dlt_offline_trace.c
phongt: @votrungchi  
clang doesn't warn that line of code. 
But I changed to use the macro DLT_OFFLINETRACE_INDEX_MAX_SIZE and there also no warning anymore.

src/console/dlt-convert.c
phongt: Changed

src/shared/dlt_offline_trace.c
votrungchi: Well, because of clang does not know this actually the size of source - I think it only warns when we use sizeof(source). It's would be good if we recalculate the remaining sizeof log_file_name (like you did at next line) after every strncat(). 
Best Regards.

phongt: Ok. changed as your suggestion.

src/console/dlt-convert.c
```diff
@@ -343,7 +343,10 @@ int main(int argc, char *argv[])
 
                         if (end == (file.counter - 1)) {
                             /* Sleep if no new message was received */
-                            sleep(0.1);
+                            struct timespec req;
+                            req.tv_sec = 0;
+                            req.tv_sec = 100000000;
```
ssugiura: `req.tv_nsec = 100000000;`

phongt: thanks. fix the mistake.

PR #144, opened 2019-06-25
dlt_user: use posix nanosleep

phongt:
For "POSIX: Replace usleep with nanosleep" as [ReleaseNotes.md](https://github.com/GENIVI/dlt-daemon/blob/master/ReleaseNotes.md#2183)

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -2798,8 +2798,12 @@ void dlt_user_trace_network_segmented_thread(void *unused)
 
         if (read < 0) {
             if (errno != EINTR) {
+                struct timespec req;
+                long sec = (DLT_USER_MQ_ERROR_RETRY_INTERVAL / 1000000);
                 dlt_vlog(LOG_WARNING, "NWTSegmented: Error while reading queue: %s\n", strerror(errno));
-                usleep(DLT_USER_MQ_ERROR_RETRY_INTERVAL);
+                req.tv_sec = sec;
+                req.tv_nsec = (DLT_USER_MQ_ERROR_RETRY_INTERVAL - sec * 1000000) * 1000;
+                nanosleep(&req, NULL);
             }
 
             continue;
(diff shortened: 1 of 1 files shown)


src/lib/dlt_user.c
```diff
@@ -2799,7 +2799,10 @@ void dlt_user_trace_network_segmented_thread(void *unused)
         if (read < 0) {
             if (errno != EINTR) {
                 dlt_vlog(LOG_WARNING, "NWTSegmented: Error while reading queue: %s\n", strerror(errno));
-                usleep(DLT_USER_MQ_ERROR_RETRY_INTERVAL);
+                struct timespec req;
```
sherkenhoff: Can you move the variable to the beginning of its scope, i.e. just after "if"?

src/lib/dlt_user.c
```diff
@@ -2817,7 +2820,10 @@ void dlt_user_trace_network_segmented_thread(void *unused)
         /* DLT_NW_TRACE_RESEND custom type is used to mark a resend */
         if(data->nw_trace_type == DLT_NW_TRACE_RESEND) {
             /* Sleep 100ms, to allow other process to read FIFO */
-            usleep(100 * 1000);
+            struct timespec req;
```
sherkenhoff: Can you move the variable to the beginning of its scope, i.e. just after "if"?

src/lib/dlt_user.c
ssugiura: Please put it before dlt_vlog(). Declaration should go to the beginning.

src/lib/dlt_user.c
```diff
@@ -2799,7 +2799,10 @@ void dlt_user_trace_network_segmented_thread(void *unused)
         if (read < 0) {
             if (errno != EINTR) {
                 dlt_vlog(LOG_WARNING, "NWTSegmented: Error while reading queue: %s\n", strerror(errno));
-                usleep(DLT_USER_MQ_ERROR_RETRY_INTERVAL);
+                struct timespec req;
+                req.tv_sec = 0;
+                req.tv_nsec = DLT_USER_MQ_ERROR_RETRY_INTERVAL * 1000;
```
ssugiura: As user might change the value of DLT_USER_MQ_ERROR_RETRY_INTERVAL to more than 1sec, this needs to be taken care.

src/lib/dlt_user.c
phongt: done

src/lib/dlt_user.c
phongt: Do  convert

PR #149, opened 2019-07-23
dlt-system-journal: fixed localtime compile error

jardous:
fixes compiler error
"error: too many arguments to function ‘localtime’"

Signed-off-by: jiripopek <Jiri.Popek@bmw.de>


ssugiura:
@jardous Thanks a lot for the fix. I would propose to change to localtime_r() instead of localtime() (and that was what I wanted to do in previous change).

PR #160, opened 2019-09-10
Propper setup and error checking of pthread_create

oncaphillis:
* Don't make pthread_t and pthread_attr_t static
* Call pthread_attr_init() and pthread_attr_destroy()
* Bail out if pthread_attr_init() or pthread_create() fails
* Check if MAX_THREADS gets exceeded


src/system/dlt-system-filetransfer.c
```diff
@@ -768,9 +768,23 @@ void filetransfer_thread(void *v_conf)
 
 void start_filetransfer(DltSystemConfiguration *conf)
 {
+    if(threads.count == MAX_THREADS)
+        DLT_LOG(dltsystem, DLT_LOG_ERROR, DLT_STRING("Could not create filetransfer thread. Out of thread slots.\n"));
```
votrungchi: Missing return here.

src/system/dlt-system-journal.c
```diff
@@ -384,12 +384,26 @@ void journal_thread(void *v_conf)
 
 void start_systemd_journal(DltSystemConfiguration *conf)
 {
+    if(threads.count == MAX_THREADS)
+        DLT_LOG(dltsystem, DLT_LOG_ERROR, DLT_STRING("Could not create journal thread. Out of thread slots.\n"));
```
votrungchi: Missing return here. Please similar check other files.

src/system/dlt-system-filetransfer.c
```diff
@@ -768,9 +768,23 @@ void filetransfer_thread(void *v_conf)
 
 void start_filetransfer(DltSystemConfiguration *conf)
 {
+    if(threads.count == MAX_THREADS)
```
votrungchi: add white space. e.g: if (

ssugiura:
@oncaphillis Your change sounds feasible. Please take care the comments from @votrungchi.
Also before merging this, I would like to ask you to do following:
- Please add signed off to your commit comment.
- Please run uncrustify as written in README.
- One line shouldn't have more than 80 columns.
Thanks.

oncaphillis:
An alternative would have been to pass NULL as the `pthread_attr_t` ptr since we seem to go with the default in all cases. 
I also would have loved to refactor and provide a genetic `create_a_thread(...)` function instead if the code duplications

oncaphillis:
uncrustify didn't seem to cut all lines down to 80 cols though

ssugiura:
@oncaphillis Thanks for your quick update. Your idea to refactor to have a generic function would be very welcome.
Unfortunately, uncrustify won't cut down to 80 columns, so you need to take care by your own.

oncaphillis:
So -- now i am going with the refactoring
* replacing all `start_XXX` with a single `start_thread(...)`
* Not using `pthread_attr_t` at all
* Hopefully cut all lines to 80c

PR #168, opened 2019-09-16
dlt-control: Wait until send buffer is empty before exiting

jardous:
By checking amount of unsent data in send buffer,
dlt-control will wait until send buffer is empty before exiting.
Thus, dlt-control ensure control message reaches to dlt-daemon.

### DIFF
--- a/src/console/dlt-control.c
+++ b/src/console/dlt-control.c
@@ -57,6 +57,7 @@
 #include <sys/stat.h>   /* for S_IRUSR, S_IWUSR, S_IRGRP, S_IROTH */
 #include <fcntl.h>      /* for open() */
 #include <sys/uio.h>    /* for writev() */
+#include <sys/ioctl.h>  /* for ioctl(), SIOCOUTQ */
 #include <string.h>     /* for open() */
 
 #include "dlt_client.h"
@@ -111,6 +112,27 @@ typedef struct {
     DltFilter filter;
 } DltReceiveData;
 
+
+#ifdef __linux__
+void wait_send_buffer_empty(int fd)
+{
+    struct timespec ts;
+    ts.tv_sec = 0;
+    ts.tv_nsec = 100000 * NANOSEC_PER_MILLISEC;
+
+    /* Wait until send buffer is empty */
+    while (1)
+    {
+        int unsent;
+        ioctl(fd, TIOCOUTQ, &unsent);
+        if (!unsent)
+        {
+            break;
+        }
+        nanosleep(&ts, NULL);
+    }
+}
+#endif
 
 /**
  * Print usage information of tool.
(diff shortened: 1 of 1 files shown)


votrungchi:
Hi @jardous, I wonder why we need to wait since normally close() always tries to complete the socket transmission.

ssugiura:
> Hi @jardous, I wonder why we need to wait since normally close() always tries to complete the socket transmission.

@votrungchi , Here's the comment from @jardous from old PR: https://github.com/GENIVI/dlt-daemon/pull/120#issuecomment-484395159

src/console/dlt-control.c
```diff
@@ -112,6 +113,27 @@ typedef struct {
 } DltReceiveData;
 
 
+void wait_send_buffer_empty(int fd)
+{
+#ifdef __linux__
```
ssugiura: This would gives you an error with unused parameter _fd_ when compiling on other PF than Linux.

src/console/dlt-control.c
```diff
@@ -112,6 +113,27 @@ typedef struct {
...
+    {
+        int unsent;
+        ioctl(fd, TIOCOUTQ, &unsent);
+        if (!unsent)
+        {
+            break;
+        }
+
+        ts.tv_sec = 0;
+        ts.tv_nsec = 100000 * 1000;
```
ssugiura: Maybe it's worth to use NANOSEC_PER_MILLISEC or NANOSEC_PER_SEC which are defined in src/console/dlt-control-common.h. Also, it's better to have these code outside of while() if it doesn't change dynamically.

votrungchi:
> 
> 
> > Hi @jardous, I wonder why we need to wait since normally close() always tries to complete the socket transmission.
> 
> @votrungchi , Here's the comment from @jardous from old PR: [#120 (comment)](https://github.com/GENIVI/dlt-daemon/pull/120#issuecomment-484395159)

Hi @ssugiura, The old PR seems different one. Other than that, my comment is about when the dlt-control exit(), it will do the close() socket which will block the process until the transmission completed if SO_LINGER option is enabled. I think it is a better solution since it is POSIX compliant.

src/console/dlt-control.c
jardous: done

src/console/dlt-control.c
jardous: done

src/console/dlt-control.c
```diff
@@ -112,6 +113,27 @@ typedef struct {
...
+    {
+        int unsent;
+        ioctl(fd, TIOCOUTQ, &unsent);
+        if (!unsent)
+        {
+            break;
+        }
+        nanosleep(&ts, NULL);
+    }
+#endif
```
ssugiura: endif should be moved to line 136

ssugiura:
@jardous Can you clarify the comment from @votrungchi ?

jardous:
@votrungchi This fix has been introduced as solution replacing missing proper between dlt-control and dlt-daemon to assure the message is processed before dlt-control process can exit. This was an issue, when the exid delay of the dlt-control was reduced

votrungchi:
@jardous I understand the problem but please consider the solution that I mentioned in previous comment, about use SO_LINGER option on the socket fd. According to [1], When enabled, a close(2) or shutdown(2) will not return until all queued messages for the socket have been successfully sent or the linger timeout has been reached. 
[1] http://man7.org/linux/man-pages/man7/socket.7.html

jardous:
@votrungchi yes, thanks for the hint. This would make sense and should be fixed while opening the socket. Let's close this pull request.

votrungchi:
@jardous You can create new pull request or just update this pull request itself. It's up to you. Thanks.

jardous:
@votrungchi unfortunately I do not have resources to do that. This is one of patch sets that we have laying around and I thought it might be useful to try to upstream it.

PR #173, opened 2019-10-02
Enable offline trace if "OfflineTraceDirectory" & "OfflineTraceOn" option is set

oncaphillis:
Instead of DLT_USER_MODE_EXTERNAL set the UserLogMode to DLT_USER_MODE_BOTH 
when "OfflineTraceDirectory" and "OfflibeTraceOn" is set in dlt.log

Some artifacts generated by  uncriustify

Signed-off-by: Sebastian Kloska <sebastian.kloska@snafu.de>

### DIFF
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -247,6 +247,7 @@ int option_file_parser(DltDaemonLocal *daemon_local)
     /* set default values for configuration */
     daemon_local->flags.sharedMemorySize = DLT_SHM_SIZE;
     daemon_local->flags.sendMessageTime = 0;
+    daemon_local->flags.offlineTraceOn = 0;
     daemon_local->flags.offlineTraceDirectory[0] = 0;
     daemon_local->flags.offlineTraceFileSize = 1000000;
     daemon_local->flags.offlineTraceMaxSize = 0;
@@ -454,6 +455,11 @@ int option_file_parser(DltDaemonLocal *daemon_local)
                     else if (strcmp(token, "SharedMemorySize") == 0)
                     {
                         daemon_local->flags.sharedMemorySize = atoi(value);
+                        /*printf("Option: %s=%s\n",token,value); */
+                    }
+                    else if (strcmp(token, "OfflineTraceOn") == 0)
+                    {
+                        daemon_local->flags.offlineTraceOn = atoi(value);
                         /*printf("Option: %s=%s\n",token,value); */
                     }
                     else if (strcmp(token, "OfflineTraceDirectory") == 0)
@@ -1000,7 +1006,8 @@ int dlt_daemon_local_init_p2(DltDaemon *daemon, DltDaemonLocal *daemon_local, in
                         daemon_local->flags.contextLogLevel,
                         daemon_local->flags.contextTraceStatus,
                         (daemon_local->flags.offlineTraceDirectory[0] !=
-                         '\0') ? DLT_USER_MODE_BOTH : DLT_USER_MODE_EXTERNAL,
+                         '\0') &&
+                        (daemon_local->flags.offlineTraceOn != 0) ? DLT_USER_MODE_BOTH : DLT_USER_MODE_EXTERNAL,
                         daemon_local->flags.enforceContextLLAndTS,
                         daemon_local->flags.vflag) == -1) {
         dlt_log(LOG_ERR, "Could not initialize daemon data\n");
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -148,8 +148,10 @@ int option_handling(DltDaemonLocal *daemon_local, int argc, char *argv[])
     opterr = 0;
 
 #ifdef DLT_SHM_ENABLE
+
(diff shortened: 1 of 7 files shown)


ssugiura:
@oncaphillis The change itself makes sense in my feeling. Could you consider to separate the commits between the original fix and the ones generated by uncrustify so that everyone knows which lines were exactly modified for the issue raised?

oncaphillis:
@ssugiura Done

oncaphillis:
Added new conf setting "OfflineTraceOn"

votrungchi:
> 
> 
> Added new conf setting "OfflineTraceOn"

Hi @oncaphillis , I wonder why we need this new option while the OfflineTraceDirectory still do the job well?

src/daemon/dlt-daemon.c
```diff
@@ -1000,7 +1007,7 @@ int dlt_daemon_local_init_p2(DltDaemon *daemon, DltDaemonLocal *daemon_local, in
                         daemon_local->flags.contextLogLevel,
                         daemon_local->flags.contextTraceStatus,
                         (daemon_local->flags.offlineTraceDirectory[0] !=
-                         '\0') ? DLT_USER_MODE_BOTH : DLT_USER_MODE_EXTERNAL,
+                         '\0') && daemon_local->flags.offlineTraceOn!=0 ? DLT_USER_MODE_BOTH : DLT_USER_MODE_EXTERNAL,
```
votrungchi: Please add the white space before and after "!="

src/daemon/dlt-daemon.c
```diff
@@ -456,6 +457,12 @@ int option_file_parser(DltDaemonLocal *daemon_local)
                         daemon_local->flags.sharedMemorySize = atoi(value);
                         /*printf("Option: %s=%s\n",token,value); */
                     }
+                    else if (strcmp(token, "OfflineTraceOn") == 0)
+                    {
+                        daemon_local->flags.offlineTraceOn = atoi(value);
+                        fprintf(stderr,"OfflineTraceOn==%d\n",daemon_local->flags.offlineTraceOn);
```
votrungchi: I think it should "OfflineTraceOn=%d\n" instead of "OfflineTraceOn==%d\n"

src/daemon/dlt-daemon.c
oncaphillis: Done

src/daemon/dlt-daemon.c
oncaphillis: ... and should be removed all togther -- debug code -- thanks for pointing out

oncaphillis:
Hi @votrungchi thanks for the review: OfflineTraceDirectory defnes the target for trace files and activates the tracing itself in one go. We like to have this behaviour in order to capture very early log messages before any dlt-client has a chance to activate offline logging. However we also need an ability to explicitly turn it off  right at the beginning without loosing the directory in order to give a dlt-client a chance to activate it later on. Does that sound ok ? Thx.

votrungchi:
Hi @oncaphillis , thanks for your detailed explanation. Could you please clarify that the dlt-client has ability to activate the OfflineTraceDirectory at runtime? If not, in order to turn off this feature without losing the directory, we could put the "#" at the beginning of the configure line. 
In addition, if we add OfflineTraceOn option, we should care about the scenario that the OfflineTraceOn is set to On but the OfflineTraceDirectory is not available (not configured yet). 
That's my opinion, so please help to consider if we really need this option. Thank you.

ssugiura:
> Hi @votrungchi thanks for the review: OfflineTraceDirectory defnes the target for trace files and activates the tracing itself in one go. We like to have this behaviour in order to capture very early log messages before any dlt-client has a chance to activate offline logging. However we also need an ability to explicitly turn it off right at the beginning without loosing the directory in order to give a dlt-client a chance to activate it later on. Does that sound ok ? Thx.

I'm not 100% sure with your use case, but I think you can use Offline Logstorage feature for that purpose.
BTW, this second proposal should be raised as a new PR, but it should be okay for this time:)

oncaphillis:
From what I can tell the "OfflineTraceDirectory" is broken. It claims to activate offline tracing but definitely doesn't 

* trace mode is set unconditionally to DLT_USER_MODE_EXTERNAL on startup (dlt-daemon.c:994) 
* Right after that we would call dlt_offline_trace_init which would actually create an open offline log file. But that never happens since the mode is DLT_USER_MODE_EXTERNAL.  
* Then all attempts to write to an nonexisting file fails @ dlt_offline_trace.c:368  
* During initilization we later read a potential dlt-runtme.cfg (dlt-daemon.c:887) file which might dictate another log mode. But we do that after potentially trying to create a file.    
* I mixed up the ability to activate offline traces via client commands with the behaviour of "OfflineLogStorageDirPath" 
* I am targeting towards a simple solution as provided by OfflineTraceDirectory with no additional files files etc. 
* My current approach is a bloated attempt to solve the problem. I close this PR and provide a simpler solution.

oncaphillis:
Please have a look @ https://github.com/GENIVI/dlt-daemon/pull/175

PR #174, opened 2019-10-07
DLT offline trace creating mutiple files even after reaching MaxTrace…

mvaradaraj:
… size when filename is index based

### DIFF
--- a/include/dlt/dlt_offline_trace.h
+++ b/include/dlt/dlt_offline_trace.h
@@ -60,7 +60,7 @@
 #include "dlt_types.h"
 
 #define DLT_OFFLINETRACE_FILENAME_BASE "dlt_offlinetrace"
-#define DLT_OFFLINETRACE_FILENAME_DELI "."
+#define DLT_OFFLINETRACE_FILENAME_DELI "_"
 #define DLT_OFFLINETRACE_FILENAME_EXT  ".dlt"
 #define DLT_OFFLINETRACE_INDEX_MAX_SIZE 10
 #define DLT_OFFLINETRACE_FILENAME_TO_COMPARE "dlt_offlinetrace_"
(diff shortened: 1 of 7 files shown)


votrungchi:
@mvaradaraj Did you verify your commit? As review your commit, I found the issue in the function dlt_offline_trace_file_name() that use sizeof(log_file_name) which is incorrect because log_file_name is the pointer. In other words, the DLT Offline trace will not work when file name is index based.
BTW, please add sign-off to your commit message.

ssugiura:
> @mvaradaraj Could you please help to confirm that the DLT Offline trace with filename is index based is working because as reviewing your commit, I found the issue in the function dlt_offline_trace_file_name() that use sizeof(log_file_name) which is incorrect because log_file_name is the pointer. In other words, the DLT Offline trace will not work when file name is index based.
> BTW, please add sign-off to your commit message.

Is it correct that the issue you mentioned here would be solved by https://github.com/GENIVI/dlt-daemon/pull/178?

ssugiura:
After discussion with @votrungchi , we agreed that we need both PRs to fix the issue.
@mvaradaraj Could you please add sign-off to your commit message like Chi requested in previous comment? Also, first line of the commit message should be less than 72 columns to give summary of the commit.

mvaradaraj:
@ssugiura I am new to the gerrit, I think I have submitted the requested changes(Signed off Information). Let me know if I missed anything..

mvaradaraj:
@ssugiura  Could you please take this commit Id 2367ea87c5a3baa372a4750bb1ed93394ce7db0a

ssugiura:
@mvaradaraj Thanks a lot. It seems that your change is covered by #186 . Could you please check if we can go with #186?

ssugiura:
@mvaradaraj I will close this PR since it should be fixed by https://github.com/GENIVI/dlt-daemon/pull/186. If there's any further concern, please reopen this PR or create a new one.

PR #175, opened 2019-10-10
Fix: Propper usage of LoggingMode in ".../dlt-runtime.cfg"

oncaphillis:
* Symptom: no offline trace file gets created
  even if runtime config LoggingMode==(2|3).

* Reason: runtime config gets read *after*
  offline logging setup.

* Solution: Read runtime config *before*
  offline logging setup.

* Make the "read runtime config" function
  set a proper default value for logging mode
  instead if the *_init_* function setting up
  the logging.

Signed-off-by: Sebastian Kloska <sebastian.kloska@snafu.de>

### DIFF
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -148,8 +148,10 @@ int option_handling(DltDaemonLocal *daemon_local, int argc, char *argv[])
     opterr = 0;
 
 #ifdef DLT_SHM_ENABLE
+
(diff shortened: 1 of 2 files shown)


votrungchi:
Hi @oncaphillis This solution look good to me. Thanks for your continuing contribute.

ssugiura:
@oncaphillis Thanks. I agree with @votrungchi . Please run uncrustify if not yet, then I can merge to master right away.

oncaphillis:
Done: Thx

Issue #179, opened 2019-10-16
buffer overrun when calling dlt_message_payload


ealmqvis:
dlt_message_payload does not consider textlength parameter before copy of data to "text"

DltReturnValue dlt_message_payload(DltMessage *msg, char *text, int textlength, int type, int verbose)

see the case:
if ((type_info & DLT_TYPE_INFO_STRG) &&
        (((type_info & DLT_TYPE_INFO_SCOD) == DLT_SCOD_ASCII) || ((type_info & DLT_TYPE_INFO_SCOD) == DLT_SCOD_UTF8))) {
in function dlt_message_argument_print

if variables "datalength" and "length" are larger than textlength still DLT_MSG_READ_STRING((text + strlen(text)), *ptr, *datalength, length); is called and length number of bytes are copied into "text" buffer beyond the textlength boundary.

This was discovered calling dlt_message_payload with a rather small text buffer. In my case size 64.

ssugiura:
@ealmqvis Could you please check this PR https://github.com/GENIVI/dlt-daemon/pull/181? This should fix the reported issue.

ealmqvis:
@ssugiura i tested the patch. unfortunately i'm not sure the patch solved everything i  still get a segmentation fault or today Illegal instruction sometimes. my buffer is on the stack and stack is corrupted...

votrungchi:
Hi @ealmqvis Thank you for your reported. Could you please help to clarify the scenario of your test which caused the segmentation fault/Illegal instruction. What hardware you used to test (Intel/ARM, 32/64 bit system)? Could you provide the coredump file, stack trace when the issue happened? Thanks.

ealmqvis:
Hi @votrungchi i'm on an arm aarch64 system (nxp ls1043)
i think the fix is solving my reported issue. but when i tested it i could see my stack was corrupted. i traced further and i can not really conclude where the corruption occurd. i have not spent more time to find the root cause, hence i have unfortunately no more debug information to give.

the call i do looks like this

TESTDATA_SIZE_LOGS = 128;
char actual[TESTDATA_SIZE_LOGS];
int verbose_flag 1
...
static int dlt_callback(DltMessage *message, void *data)
{
...
result = dlt_message_payload(message, actual, TESTDATA_SIZE_LOGS, DLT_OUTPUT_ASCII, verbose_flag);
...
}

votrungchi:
[...]

PR #181, opened 2019-10-17
Avoided Seg fault in dlt_message_payload (#179)

votrungchi:
Description:

 Avoided Seg fault by adding boundary check before buffer access.

Signed-off-by: Ravi Sankar P <ponnurangamravi.sankar@in.bosch.com>


votrungchi:
> 
> 
> Perhaps `strlen` could be called only once.

Done with the 6abe94c commit. Thank you.

ealmqvis:
thanks for the fix. looks good

PR #184, opened 2019-11-04
libdlt: Fix message lost issue at exit

jardous:
If messages are stored to buffer in user library and dlt_unregister_app() is
called before exit, message lost occurred due to missing appID.

Change-Id: I1aef367fef4e14027114cf8c7e58e65fda6763eb

### DIFF
--- a/include/dlt/dlt_user.h
+++ b/include/dlt/dlt_user.h
@@ -190,6 +190,7 @@ typedef struct
 {
     char ecuID[DLT_ID_SIZE];                   /**< ECU ID */
     char appID[DLT_ID_SIZE];                   /**< Application ID */
+    char appID_atexit[DLT_ID_SIZE];            /**< Application ID for atexit */
     int dlt_log_handle;                        /**< Handle to fifo of dlt daemon */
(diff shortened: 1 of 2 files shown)


ssugiura:
@jardous You can use DLT_UNREGISTER_APP_FLUSH_BUFFERED_LOGS instead of DLT_UNREGISTER_APP to flush all data in internal buffer, then we don't need to take care in atexit_handler. Could you please verify whether this works?

ssugiura:
@jardous This is a friendly reminder.

jardous:
@ssugiura yes, I got finally confirmed, that this will work for us. But seems like it has to be handled separately by each client though. Thank you.

ssugiura:
@jardous Yes, it needs to be configured by each application. DLT_UNREGISTER_APP_FLUSH_BUFFERED_LOGS would definitely make sure that all data in the internal buffers are sent to the daemon, but on the other hand, it would take some time before unregistration. So it's up to each application which macro should be used.
So, do you want to close this PR or would like to modify some codes?

jardous:
abandoned - cleaner solution proposed

PR #187, opened 2019-11-11
libdlt: add an interface to get the status of application registering

zeerd:
I am developing a middleware with a single library. And I am using the dlt to output the logs.
But I find some of the executables that using my library was not using dlt, it means none will register the application, and my library could not output any logs.
But if I register the application in my library, my application id will override the application id of the executable who using dlt.

So, I did this change to make the register-application only work once.
If the excutable registered the application, my library 's registering will be ignored.
Only thing might change is, the executable need to register the application as soon as possible, early before any libraries.

Signed-off-by: Charles Chan <charles@zeerd.com>

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -366,7 +366,6 @@ DltReturnValue dlt_init(void)
     dlt_user.dlt_is_file = 0;
     dlt_user.overflow = 0;
     dlt_user.overflow_counter = 0;
-    dlt_user.app_reference = 0;
 #ifdef DLT_SHM_ENABLE
     memset(&(dlt_user.dlt_shm), 0, sizeof(DltShm));
 
@@ -1002,8 +1001,6 @@ DltReturnValue dlt_register_app(const char *apid, const char *description)
         return DLT_RETURN_OK;
     }
 
-    dlt_user.app_reference++;
-
     DLT_SEM_LOCK();
 
     /* Store locally application id and application description */
@@ -1338,10 +1335,6 @@ DltReturnValue dlt_unregister_app(void)
     dlt_user.application_description = NULL;
 
     DLT_SEM_FREE();
-
-    if(dlt_user.app_reference > 0) {
-        dlt_user.app_reference--;
-    }
 
     return ret;
 }
@@ -4685,5 +4678,5 @@ DltReturnValue dlt_user_log_out_error_handling(void *ptr1, size_t len1, void *pt
 
 int dlt_is_app_registered(void)
 {
-    return (dlt_user.app_reference > 0) ? 1 : 0;
+    return (dlt_user.appID[0] == '\0') ? 0 : 1;
 }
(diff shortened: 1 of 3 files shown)


ssugiura:
@zeerd Thanks for your proposal. But from your use case, I feel your library should take care to decide whether to call dlt_register_app() or not.

zeerd:
But, I have no method to know if the executable had registered or not.

votrungchi:
@zeerd You're right. Please add the interface to get the status of dlt_user.reference flag (maybe you should think about the meaning name and its value should 0 or 1 only).
@ssugiura What do you think?

zeerd:
@votrungchi In such a situation, I think I can add a get-function to read the appID, and I need no new variable in struct DltUser.

Maybe like:
const char* dlt_get_app_id(void)
{
    return dlt_user.appID;
}

I will register an application if the length of the returned value is 0.

zeerd:
I changed my mind anded add a new interface to return 0 and 1.

src/lib/dlt_user.c
```diff
@@ -1002,46 +1002,38 @@ DltReturnValue dlt_register_app(const char *apid, const char *description)
         return DLT_RETURN_OK;
     }
 
-    dlt_user.reference++;
+    dlt_user.app_reference++;
```
votrungchi: I think you should set it to 1 instead of increase (if you still want to increase, you have to set it to 0 instead of decrease in case of dlt_unregister_app) and only after dlt_user_log_send_register_application() return OK at line 1033 to make sure the application is registered successfully.

src/lib/dlt_user.c
```diff
@@ -1332,28 +1324,23 @@ DltReturnValue dlt_unregister_app(void)
...
-        DLT_SEM_FREE();
+    DLT_SEM_FREE();
 
-        dlt_vlog(LOG_DEBUG, "%s done\n", __FUNCTION__);
-    }
-    else {
-        dlt_vlog(LOG_DEBUG, "%s ignored for reference=%d\n", __FUNCTION__, dlt_user.reference);
-        dlt_user.reference--;
+    if(dlt_user.app_reference > 0) {
+        dlt_user.app_reference--;
```
votrungchi: As my comment above, you should set it to 0.

zeerd:
By your suggest, I think the new variable is no use any more. We could check the appID[0] directly.

votrungchi:
> 
> 
> By your suggest, I think the new variable is no use any more. We could check the appID[0] directly.

@zeerd If check the appID[0] could solve your problem. We already have the commit locally that provided the interface to support get appID and will create a pull request to upstream.

zeerd:
OK, thank you.

 I will close this pull request, then. And waiting for your release.

ssugiura:
@zeerd @votrungchi Sorry that I couldn't update for long time and thanks a lot for your discussion. I will merge #188 as concluded in this discussion.

PR #188, opened 2019-11-13
Provide GET_APPID macro (#187)

votrungchi:

### DIFF
--- a/include/dlt/dlt_user.h
+++ b/include/dlt/dlt_user.h
@@ -537,6 +537,13 @@ DltReturnValue dlt_unregister_app(void);
 DltReturnValue dlt_unregister_app_flush_buffered_logs(void);
 
 /**
+ * Get the application id
+ * @param four byte long character array to store the application id
+ * @return Value from DltReturnValue enum
+ */
+DltReturnValue dlt_get_appid(char *appid);
+
+/**
  * Register a context in the daemon.
  * This function has to be called before first usage of the context.
  * @param handle pointer to an object containing information about one special logging context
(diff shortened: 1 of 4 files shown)


src/lib/dlt_user.c
```diff
@@ -414,6 +414,18 @@ DltReturnValue dlt_init(void)
...
 
+void dlt_get_appid(char * appid)
+{
+    if(appid != NULL)
+    {
+        strncpy(appid, dlt_user.appID, 4);
+    }
+    else
+    {
+        dlt_log(LOG_ERR, "Invalid parameter.\n");
```
ssugiura: From #187 's use case, it's better to return error value instead of just printing error message.

include/dlt/dlt_user.h
```diff
@@ -536,6 +536,13 @@ DltReturnValue dlt_unregister_app(void);
  */
 DltReturnValue dlt_unregister_app_flush_buffered_logs(void);
 
+/**
+ * provide the registered appid
+ * This function has to be called before registering an application.
+ * @param character pointer of minimum four bytes to provide appid registered
+ */
+void dlt_get_appid(char * appid);
```
ssugiura: Uncrustify will do the cleanup here:)

src/lib/dlt_user.c
votrungchi: I think #187's user case care about the value of appid but not the return value. In case of error, appid will be NULL or empty. Isn't it?

ssugiura: Actually, it depends what the caller set to the parameter since this function won't do any initialization...

PR #191, opened 2019-11-23
Change DLT_CHECK_RCV_DATA_SIZE macro to an internal function

### LINKED ISSUE #190: Use of DLT_CHECK_RCV_DATA_SIZE() is non-standard C++
The use of `DLT_CHECK_RCV_DATA_SIZE()` in a if condition as in [`src/daemon/dlt_daemon_client.c`](https://github.com/GENIVI/dlt-daemon/blob/346781ff633aee48eee5700a0b379d97cf86c935/src/daemon/dlt_daemon_client.c) and [`src/gateway/dlt_gateway.c`](https://github.com/GENIVI/dlt-daemon/blob/346781ff633aee48eee5700a0b379d97cf86c935/src/gateway/dlt_gateway.c) is non-standard C++ code. The if condition must be either an [expression or a declaration of a single non-array variable](https://en.cppreference.com/w/cpp/language/if).

Nevertheless, many compilers seem to accept this code. Try this yourself on [godbolt](https://godbolt.org/z/rZRrU9). However, compiling with `-pedantic` results even with gcc in the following warning:
```
<source>:2:1: warning: ISO C++ forbids braced-groups within expressions [-Wpedantic]
```
MSVC also fails to compile this code.

The use of non-standard C++ makes it more difficult to port this DLT implementation to other platforms like Windows and proprietary embedded.

A possible solution could be to change this macro to a static inline function.

sebastianlipponer:
fixes #190

Changing the `DLT_CHECK_RCV_DATA_SIZE` macro into a static inline function requires to include `syslog.h` (due to the use of `LOG_WARNING`) and to declare `dlt_vlog()` earlier. Therefore, I think it's acutally best to just change this macro into an internal function.

### DIFF
--- a/include/dlt/dlt_common.h
+++ b/include/dlt/dlt_common.h
@@ -218,17 +218,6 @@ enum {
 #   define GET_LOG_INFO_LENGTH 13
 #   define SERVICE_OPT_LENGTH 3
 
-/* checks if received size is big enough for expected data */
-#   define DLT_CHECK_RCV_DATA_SIZE(received, required) \
-    ({ \
-        int _ret = DLT_RETURN_OK; \
-        if (((int)received - (int)required) < 0) { \
-            dlt_vlog(LOG_WARNING, "%s: Received data not complete\n", __func__); \
-            _ret = DLT_RETURN_ERROR; \
-        } \
-        _ret; \
-    })
-
 /**
  * Get the size of extra header parameters, depends on htyp.
  */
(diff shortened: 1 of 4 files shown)


ssugiura:
@sebastianlipponer The change itself looks good to me. Two things to ask to your commit message:
- Please explain what was the problem and how you solved it so that everyone can understand what you fixed with this commit.
- Please add signed-off
Once you've updated I can merge the commit right away.

sebastianlipponer:
Thanks for your comments. Update done.

PR #194, opened 2019-12-01
fix some warnings from coverity scan

phongt:
For fixing  resource leak and unint  warning by coverity

### DIFF
--- a/src/console/logstorage/dlt-logstorage-list.c
+++ b/src/console/logstorage/dlt-logstorage-list.c
@@ -165,6 +165,8 @@ int logstorage_store_dev_info(const char *node, const char *path)
 
     if (ptr->mnt_point == NULL) {
         pr_error("memory allocation failed for mnt_point\n");
+        free(ptr);
+        ptr = NULL;
         return -1;
     }
 
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -650,6 +650,7 @@ int option_file_parser(DltDaemonLocal *daemon_local)
 
                             if (daemon_local->flags.ipNodes == NULL) {
                                 dlt_vlog(LOG_ERR, "Could not allocate for IP list\n");
+                                fclose(pFile);
                                 return -1;
                             }
                             else {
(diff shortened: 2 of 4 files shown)


src/console/logstorage/dlt-logstorage-list.c
```diff
@@ -165,6 +165,7 @@ int logstorage_store_dev_info(const char *node, const char *path)
 
     if (ptr->mnt_point == NULL) {
         pr_error("memory allocation failed for mnt_point\n");
+	free(ptr);
```
ssugiura: Please avoid to use tab and use 4 spaces instead. Uncrustify should do the job for this. Please refer to README which version to use.
Also, NULL should be set to ptr after free.

phongt: >Please avoid to use tab and use 4 spaces instead. Uncrustify should do the job for this. Please refer to README which version to use.

Checked with unncrustify, no finding with this pr. Somehow there are also some not good style not in changing code with this pr. Another PR ;-) ?

Eg:
```
diff --git a/src/daemon/dlt-daemon.c b/src/daemon/dlt-daemon.c
index b2d3668..d6f013b 100644
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -1205,13 +1205,13 @@ static int dlt_daemon_init_fifo(DltDaemonLocal *daemon_local)
         return -1;
     } /* if */
 
-    if (daemon_local->daemonFifoSize != 0) {
+    if (daemon_local->daemonFifoSize != 0)
         /* Set Daemon FIFO size */
         if (fcntl(fd, F_SETPIPE_SZ, daemon_local->daemonFifoSize) == -1)
             dlt_vlog(LOG_ERR, "set FIFO size error: %s\n", strerror(errno));
-    }
 
     /* Get Daemon FIFO size */
+
     if ((fifo_size = fcntl(fd, F_GETPIPE_SZ, 0)) == -1)
         dlt_vlog(LOG_ERR, "get FIFO size error: %s\n", strerror(errno));
     else
```

>Also, NULL should be set to ptr after free.

Done.

phongt: @ssugiura if you interested in Coverity Scan. youi can take a look [here](https://scan.coverity.com/projects/phongt-dlt-daemon).

ssugiura: @phongt Wow, unexpected:) Let us improve uncrustify configuration and we will take care. Thanks for pointing out. And thanks for sharing us the coverity scan.

PR #205, opened 2020-02-01
doc/dlt_for_developers.md: Fix wrong include directive

sebastianlipponer:
Fixes the problem that users are not able to use DLT with the include directive `#include <dlt/dlt.h>` that is stated in the developer guide (see `doc/dlt_for_developers.md`).

This is required in order to build other GENIVI projects like (vsomeip, capicxx-core-runtime, ...) without any problems since they all seem to use the include directive mentioned above.

### DIFF
--- a/doc/dlt_for_developers.md
+++ b/doc/dlt_for_developers.md
@@ -27,7 +27,7 @@ minimal code example. Detailed information about the API can be found later in
 this document.
 
 ```
-#include <dlt/dlt.h>
+#include <dlt.h>
 
 DLT_DECLARE_CONTEXT(ctx); /* declare context */
 


ssugiura:
@sebastianlipponer I guess what's written in `doc/dlt_for_developers.md` is not correct. The old document says that it should be `#include <dlt.h>`. Also, I don't want to change `automotive-dlt.pc.in` which should be already used by other people.
So I would suggest not to modify `automotive-dlt.pc.in` but correct the document, and change the header.
BTW, I have never heard of this issue in GENIVI projects before and our package config and their headers are not changed for quite a long time... was this happening before?

sebastianlipponer:
Where do I find the old document? `doc/dlt_for_developers.md` came into existence with this commit 0d0c746 and the relevant parts were never changed.

For example [vsomeip](https://github.com/GENIVI/vsomeip/blob/9fb9beecadf52083599302fa8ddee7efbec64a39/implementation/logging/src/dlt_sink_backend.cpp#L9), [capicxx-core-runtime](https://github.com/GENIVI/capicxx-core-runtime/blob/2fd0625d21f1fa8e6a3adfc89ce9f381a4d33990/include/CommonAPI/LoggerImpl.hpp#L10), [ramses](https://github.com/GENIVI/ramses/blob/0256066bd77d9b6d6ccd060231d7f75fb5bbaab7/framework/DltLogAppender/src/DltAdapterImpl/DltAdapterImpl.cpp#L15) use this way of including DLT.

I guess meta-ivi does not configure those components to actually make use of DLT. But compiling vsomeip for example with `USE_DLT` does not work with the current `automotive-dlt.pc`.

In practice both include directives seem to be in use. Even across GENIVI projects. A mere change of the documentation has probably a lower impact than my proposal.

sebastianlipponer:
@ssugiura I updated the commit to change the documentation as you have suggested.

ssugiura:
@sebastianlipponer Sorry for my late reply. Actually the same was done on old documents (doc/dlt_cheatsheet.txt and doc/dlt_user_manual.txt) which was already removed. Please refer to commit 9e101ff434230a95fb8f4fd33dc48f4970496d1c.
And thank you for the update. I will merge it to master.

PR #214, opened 2020-03-12
FIX: prevent usage of uninitialized message queue handle

RudiSV:
Signed-off-by: Rudolf Svanda <rudolf.svanda@globallogic.com>

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -2878,8 +2878,10 @@ void dlt_user_trace_network_segmented_thread(void *unused)
         /* Wait until message queue is initialized */
         dlt_lock_mutex(&mq_mutex);
 
-        if (dlt_user.dlt_segmented_queue_read_handle < 0)
+        while (dlt_user.dlt_segmented_queue_read_handle < 0)
+        {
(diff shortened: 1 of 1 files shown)


sebastianlipponer:
Note, that for `pthread_cond_wait()` spurious [wakeups may occur](https://linux.die.net/man/3/pthread_cond_wait). The current code does not account for this. Therefore, `dlt_segmented_queue_read_handle` might be uninitialized after `pthread_cond_wait` returns.

I think your change fixes the problem. However, I find your code hard to understand since it relies on the outer `while(1)` loop. What about replacing both `if()` conditions with a separate `while()` loop instead?

RudiSV:
Fix does not target "spurious wakeups" even if it fix it too.
Problem occurs always if application exists and dlt_user.dlt_segmented_queue_read_handle is not initialized (message queue is not created)

Commit is updated.

ssugiura:
@RudiSV I'm very sorry for replying late. The change itself looks good. Before merging to master, could you summarize the issue you faced and how you solved it in commit message so that others know what this commit improves? Thank you!

RudiSV:
Hope it's better now.

ssugiura:
@RudiSV Yes, it is! Thank you.

PR #215, opened 2020-03-13
dlt_common: Fix buffer overflow in dlt_buffer_get

ricleite:
A DLT lib user can create a message large enough that `dlt_user.resend_buffer` is not large enough, and thus it will overflow.

Signed-off-by: Ricardo Leite <rleite@dcc.fc.up.pt>

Seen some crashes where this turned out to be the root cause.

### DIFF
--- a/src/shared/dlt_common.c
+++ b/src/shared/dlt_common.c
@@ -2643,11 +2643,13 @@ int dlt_buffer_get(DltBuffer *buf, unsigned char *data, int max_size, int delete
     }
 
     /* third check size */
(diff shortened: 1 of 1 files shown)


jardous:
anybody willing to test and merge?

ssugiura:
@ricleite Sorry for my late response. The change seems to be reasonable to me. Could you improve the commit message by referring to https://at.projects.genivi.org/wiki/display/PROJ/Rules+for+Commit+Messages?

ricleite:
Will do, thanks!

PR #216, opened 2020-03-13
Offline Logstorage: allow both LogAppName and ContextName to be ".*"

fmntf:
In [Offline Logstorage documentation](https://github.com/GENIVI/dlt-daemon/blob/master/doc/dlt_offline_logstorage.md) it is stated that `LogAppName` and `ContextName` may be set to `.*` to match against any value.

However, if both options are set to `.*`, the wildcard is applied only to application ID.
I think this is a mistake (the documentation does not specify that the wildcard cannot be applied to both fields simultaneously). This patch allows to specify both fields as wildcard.

Test configuration:
```
root@mx8:/media/log# cat dlt_logstorage.conf 
[FILTER1]
LogAppName=.*
ContextName=.*
LogLevel=DLT_LOG_DEBUG
File=ECU2
FileSize=102400
NOFiles=10
EcuID=ECU2
```

### DIFF
--- a/src/offlinelogstorage/dlt_offline_logstorage.c
+++ b/src/offlinelogstorage/dlt_offline_logstorage.c
@@ -473,6 +473,29 @@ DLT_STATIC int dlt_logstorage_get_keys_list(char *ids, char *sep, char **list,
 }
 
 /**
+ * dlt_logstorage_create_keys_any
+ *
+ * Prepares keys without app ID and context ID, will use ecuid if provided
+ * (ecuid\:\:) or (\:\:)
+ *
+ * @param ecuid          ECU ID
+ * @param ctid           Context ID
+ * @param key            Prepared key stored here
+ * @return               None
+ */
+DLT_STATIC void dlt_logstorage_create_keys_any(char *ecuid, char *ctid,
+                                                     char *key)
+{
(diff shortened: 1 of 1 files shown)


ssugiura:
@fmntf I'm very sorry for my late reply. Instead of introducing new function, you can add the condition for dlt_logstorage_create_keys_only_ecu since EcuID is mandatory in this case. If you are not in too hurry, I can share the change which we already have done locally.

fmntf:
Hi, no hurry, take your time! Thank you!

ssugiura:
@fmntf All right then. I will inform you once it's available. Thank you for your patience!

ssugiura:
@fmntf I'm very sorry that we couldn't update very soon. With latest commits we've merged, it should be possible to use wildcards to both app and context name if Ecu ID is specified. Please verify with the latest one.
Thank you!

fmntf:
Hi @ssugiura, thank you for your update. I cannot test the master now since the team is currently busy. I will do it asap.
Anyway, is there a schedule for the next stable release?
Thank you,
Francesco

ssugiura:
@fmntf Thanks. We plan to release new 2.18.5 version within 2 weeks.

PR #219, opened 2020-03-23
timers: Replace timerfd_create() with POSIX api

sebastianlipponer:
The current timer implementation relies on `timerfd_create()` which is only available under Linux. Hence, using this DLT implementation on a POSIX compliant operating system other than Linux results in broken timers.

This change fixes this problem using a timer implementation based on `clock_gettime()`.

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -46,9 +46,6 @@
 #include <poll.h>
 
 #include <limits.h>
-#ifdef linux
-#   include <sys/prctl.h>
-#endif
 
 #include <sys/types.h> /* needed for getpid() */
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


ssugiura:
@sebastianlipponer I'm very sorry that I could not update for long time. Actually I wanted to do some performance measurement since we also have a modification locally for the same feature. As soon as the measurement is done, I will update. Thank you for your patience.

sebastianlipponer:
@ssugiura Thanks for your feedback. How did you replace timerfd?

ssugiura:
@sebastianlipponer We created a new thread and notify to main thread via unamed pipes so that the implementation would follow very similar to timerfd implementation.

ssugiura:
@sebastianlipponer 
Sorry that it took so long to verify. Finally I pushed our solution into devel branch (not in master yet). Could you check if this is also fine?
https://github.com/GENIVI/dlt-daemon/pull/279/commits/53a495ffdb3cda5dc8d6d143fd0e89b0d00816c4

sebastianlipponer:
@ssugiura Thanks for your feedback. Do you know whether the new implementation is known to work also with Green Hills Integrity?

ssugiura:
@sebastianlipponer Unfortunately we've never tried on GHS

sebastianlipponer:
@ssugiura I don't know when I will be able to test this. If you tested on QNX then I suggest to merge it.

PR #221, opened 2020-04-07
Added slog2 to DLT adapter

karthikshanmugam:
Typically QNX drivers and stack will use slogger2 to log the debug messages, this adapter will route the slogger2 messags to DLT so that the slogger2 logs can be streamed to DLT viewer in the PC and also use the DLT offline log storage feature to store the logs in the persistent memory

### DIFF
--- a/src/adaptor/CMakeLists.txt
+++ b/src/adaptor/CMakeLists.txt
@@ -27,4 +27,13 @@ install(TARGETS dlt-adaptor-stdin dlt-adaptor-udp
 	RUNTIME DESTINATION bin
 	COMPONENT base)
 
-
+if(CMAKE_SYSTEM_NAME MATCHES QNX)
+	set(dlt_adaptor_slog2_SRCS dlt-adaptor-slog2.c)
+	add_executable(dlt-adaptor-slog2 ${dlt_adaptor_slog2_SRCS})
+	target_link_libraries(dlt-adaptor-slog2 dlt slog2parse)
+	set_target_properties(dlt-adaptor-slog2 PROPERTIES LINKER_LANGUAGE C)
+	
+	install(TARGETS dlt-adaptor-slog2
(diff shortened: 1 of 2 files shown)


ssugiura:
@karthikshanmugam I would like to raise some points:
- Any reason both DEBUG1 and DEBUG2 have the same log level as DLT_LOG_VERBOSE? It should be possible to filter in more detail, so using two different levels sounds more reasonable, like DLT_LOG_DEBUG for DEBUG1 and DLT_LOG_VERBOSE for DEBUG2.
- Thinking of production use case, I propose to have a new configuration file instead of command line option to define application ID and context ID, something like dlt-system.conf.
Other than that, the implementation looks good to me.

karthikshanmugam:
@ssugiura  agreed, I will update on the log levels.
On the configuration file, I will improve on top of this when I get more, need more info on this.

karthikshanmugam:
@ssugiura Should I make a new MR or update this one?

ssugiura:
@karthikshanmugam Thanks for your quick update.
For configuration file, you can refer to dlt-system.conf for your example where they set App ID and Cxt ID there and dlt-system would read them to set the user's expected variable. Also default log level can be set there too, then user can start the adaptor without any options.
BTW, log level should be defined in number like in other command line tools e.g. dlt-control. Then you can just use defined enums to map them.
Please update this PR instead of creating a new one. Thank you!

karthikshanmugam:
@ssugiura  Do you need the conf file changes before merging this MR ?

ssugiura:
@karthikshanmugam Yes, it should be better.

ssugiura:
@karthikshanmugam I've merged the one which does the similar. Please check if it supports the feature you need. Thank you! https://github.com/GENIVI/dlt-daemon/commit/c78d08d63f47ec2baa08c20f3068adbe8c3cc0ca

PR #223, opened 2020-04-21
Add sized string functions

mawillers:
This use case is common in C++ code nowadays, due to the usage of e.g.
std::string_view, which does not provide a function for retrieving a
null-terminated C-string, but which provides quick access to the string
length via std::string_view::size().

Such use cases can now avoid an entirely unnecessary strlen() call.

### DIFF
--- a/doc/dlt_for_developers.md
+++ b/doc/dlt_for_developers.md
@@ -566,6 +566,35 @@ DLT_VERBOSE_MODE();
 DLT_NONVERBOSE_MODE();
 ```
 
+#### String arguments
+
+For string arguments, you can choose between ASCII and UTF-8 encoding. This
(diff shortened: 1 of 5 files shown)


ssugiura:
@mawillers Thank you for the suggestion. The idea sounds reasonable to me. A few points to add:
- Most of the interfaces are defined as macro in include/dlt/dlt_user_macros.h. I guess the same shall be done.
- Please update some documents e.g. doc/dlt_for_developers.md
- Please add test case too.

mawillers:
@ssugiura, happy to do so. Please have a look again.

ssugiura:
@mawillers Looks good. Thank you for your update!

PR #231, opened 2020-06-25
Add QNX slogger2 adaptor

yyatsenk:
Here is the work which was done for Volvo needs but it can be useful not only for Volvo I suppose. Here is the change which makes it possible to redirect output of slogger2 (QNX specific logger) tool to dlt. It was tested on QNX-run systems. So please review the change and It would be cool to get any comments or criticism.


ssugiura:
@yyatsenk I've merged the one which does the similar. Please check if it supports the feature you need. Thank you! https://github.com/GENIVI/dlt-daemon/commit/c78d08d63f47ec2baa08c20f3068adbe8c3cc0ca

PR #240, opened 2020-08-03
Implement DLTClient for UDP multicast

joaodasousa:
1 - change line:
        #define MULTICAST_GROUP_ADDRESS "239.1.2.3" on file dlt-example-multicast-client.c
2 - Compile and run dlt-example-multicast-client example application


ssugiura:
@joaodasousa Apart from comments in the changes, I would like to ask you two things:
- Please explain what was the problem and how you solved/improved it so that everyone can understand what you fixed with this commit.
- Please add signed-off
Thank you!

src/examples/CMakeLists.txt
```diff
@@ -42,9 +42,13 @@ if(WITH_UDP_CONNECTION)
...
 			RUNTIME DESTINATION bin
 			COMPONENT base)
-
 endif(WITH_UDP_CONNECTION)
 
-install(TARGETS dlt-example-user dlt-example-user-func dlt-example-user-common-api dlt-example-filetransfer
+set(dlt_example_multicast_client_SRCS dlt-example-multicast-client.c)
+add_executable(dlt-example-multicast-client ${dlt_example_multicast_client_SRCS})
+target_link_libraries(dlt-example-multicast-client dlt)
+set_target_properties(dlt-example-multicast-client PROPERTIES LINKER_LANGUAGE C)
```
ssugiura, line 50: Add these lines inside `if(WITH_UDP_CONNECTION)`.

src/examples/dlt-example-multicast-client.c
```diff
@@ -0,0 +1,102 @@
...
+#include <stdio.h>
+#include <syslog.h>
+#include <time.h>
+#include <sys/time.h>
+
+#include "dlt_client.h"
+#include "dlt_client_cfg.h"
+#include "dlt_types.h"
+
+#define MULTICAST_GROUP_ADDRESS "239.1.2.3"
```
ssugiura, line 33: Can we align this address to default defined in `dlt.conf`?

src/lib/dlt_client.c
```diff
@@ -153,16 +154,17 @@ DltReturnValue dlt_client_init(DltClient *client, int verbose)
         dlt_vlog(LOG_INFO,
                  "Init dlt client struct with default port: %hu.\n",
                  servPort);
-
     return dlt_client_init_port(client, servPort, verbose);
 }
 
 DltReturnValue dlt_client_connect(DltClient *client, int verbose)
 {
+    const int yes = 1;
```
ssugiura, line 162: As this value is used only if DLT_CLIENT_MODE_UDP_MULTICAST, I propose to define this line inside there. Same for line 167

src/shared/dlt_common.c
```diff
@@ -2035,12 +2037,23 @@ int dlt_receiver_receive(DltReceiver *receiver, DltReceiverType from_src)
...
+            /* wait for data from UDP socket */
+            addrlen = sizeof(receiver->addr);
+            receiver->bytesRcvd = recvfrom(receiver->fd,
+                                           receiver->buf + receiver->lastBytesRcvd,
+                                           receiver->buffersize - receiver->lastBytesRcvd,
+                                           0,
+                                           (struct sockaddr *)&(receiver->addr),
+                                           &addrlen);
+        }
+
```
ssugiura, line 2056: Please correct indention here

src/examples/CMakeLists.txt
joaodasousa: Agree

src/examples/dlt-example-multicast-client.c
joaodasousa: Agree

src/shared/dlt_common.c
joaodasousa: Agree

src/lib/dlt_client.c
joaodasousa: I don't agree with this comment.
Defining variables after labels (DLT_CLIENT_MODE_UDP_MULTICAST) is not valid in C. Although workarounds exists, in my opinion they are kinda ugly :)

ssugiura:
@joaodasousa I took a deeper look and have an idea that we can merge the code in src/examples/dlt-example-multicast-client.c to src/console/dlt-receive.c so that we would have one common DLT log viewer client. What do you think?

joaodasousa:
> @joaodasousa I took a deeper look and have an idea that we can merge the code in src/examples/dlt-example-multicast-client.c to src/console/dlt-receive.c so that we would have one common DLT log viewer client. What do you think?

Makes perfect sense. I have done as suggested

PR #247, opened 2020-09-12
Fix installation for when built with CMAKE_INSTALL_PREFIX=/

martin-ejdestig:
Want to do this when cross compiling for a aarch64le QNX platform. Without this fix, the following error occurs when installing:

```shell
$ cmake -Bbuild -S. -DCMAKE_INSTALL_PREFIX=/
...
$ cmake --build build --parallel $(JLEVEL) --target install DESTDIR=${BSP_ROOT}/install/aarch64le
...
CMake Error at src/daemon/cmake_install.cmake:70 (file):
  file called with network path DESTINATION.  This does not make sense when
  using DESTDIR.  Specify local absolute path or remove DESTDIR environment
  variable.

  DESTINATION=

  //etc
...
```

Fix by using CMAKE_INSTALL_FULL_SYSCONFDIR from GNUInstallDirs.cmake to set CONFIGURATION_FILES_DIR and remove special casing on CMAKE_INSTALL_PREFIX. Could add another special case but GNUInstallDirs.cmake is already included so might as well use it.

Might make sense to remove CONFIGURATION_FILES_DIR and use CMAKE_INSTALL_(FULL_)SYSCONFDIR directly everywhere, but not sure if CONFIGURATION_FILES_DIR is set from outside by users. Could cleanup in a separate commit if wanted.

There are also other uses of CMAKE_INSTALL_PREFIX that results in //..., but not anywhere that is relevant for QNX. Perhaps it makes sense to clean this up in a separate commit as well.


martin-ejdestig:
Of course reproducible without this fix when building for e.g. Linux as well with `CMAKE_INSTALL_PREFIX=/` and `DESTDIR=<install dir>`.

martin-ejdestig:
Decided to install under `/usr` instead so no longer need this patch. (One reason being that CMake authors have misinterpreted the GNU documentation and prepends `/usr` for some paths, see GNUInstallDirs documentation, even though e.g. autotools with `--prefix=/` do not. Result wiht `prefix=/` in many CMake projects is that things may or may not end up in `/usr` depending on how projects use CMake...).

But, I think it makes sense to use `CMAKE_INSTALL_FULL_SYSCONFDIR` regardless. If patch is still wanted, I can simplify the commit message a bit perhaps?

martin-ejdestig:
Updated commit message to instead focus on simplification rather than building with `CMAKE_INSTALL_PREFIX=/`.

PR #256, opened 2020-10-25
dlt-common.c: ensure null terminated string

mwarning:
make sure strncpy writes a null terminated string

Signed-off-by: Moritz Warning <moritzwarning@web.de>

### DIFF
--- a/src/shared/dlt_common.c
+++ b/src/shared/dlt_common.c
@@ -3844,7 +3844,7 @@ int dlt_set_loginfo_parse_service_id(char *resp_text,
 
     /* ascii type, syntax is 'get_log_info, ..' */
     /* check target id */
-    strncpy(get_log_info_tag, "get_log_info", strlen("get_log_info"));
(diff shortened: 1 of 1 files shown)


mwarning:
Demonstration for the second stncpy:
```
#include <stdio.h>
#include <string.h>

#define GET_LOG_INFO_LENGTH 13

int
main(int argc, char *argv[])
{
  char get_log_info_tag[GET_LOG_INFO_LENGTH];

  // initializes with 'X'
  memset(get_log_info_tag, 'X', sizeof(get_log_info_tag));

  // prints XXXXXXXXXXXXX
  printf("%*s\n", sizeof(get_log_info_tag), get_log_info_tag);
   
  strncpy(get_log_info_tag, "get_log_info", strlen("get_log_info"));

  // prints get_log_infoX
  printf("%s\n", get_log_info_tag);

  return 0;
}
```

mwarning:
Another fix would be to 0 initialize the array instead.

PR #258, opened 2020-10-25
dlt_client.c: remove misleading error message

mwarning:
Several IP address might be tested for connect.
Only output an error if no address succeeds at all.

Signed-off-by: Moritz Warning <moritzwarning@web.de>

### DIFF
--- a/src/lib/dlt_client.c
+++ b/src/lib/dlt_client.c
@@ -160,6 +160,7 @@ DltReturnValue dlt_client_init(DltClient *client, int verbose)
 DltReturnValue dlt_client_connect(DltClient *client, int verbose)
 {
     const int yes = 1;
+    int connect_errno = 0;
     char portnumbuffer[33];
     struct addrinfo hints, *servinfo, *p;
     struct sockaddr_un addr;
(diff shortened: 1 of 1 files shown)


mwarning:
It is slightly irritating since the connection failed message is seen and then the connection succeeds.

src/lib/dlt_client.c
```diff
@@ -199,7 +197,7 @@ DltReturnValue dlt_client_connect(DltClient *client, int verbose)
         freeaddrinfo(servinfo);
 
         if (p == NULL) {
-            dlt_log(LOG_ERR, "ERROR: failed to connect.\n");
+            dlt_vlog(LOG_ERR, "ERROR: failed to connect! %s\n", strerror(errno));
```
ssugiura, line 200: This line should return errno from connect() system call. Means if some lines are added in future between the two lines, the errno might be updated and this log doesn't return expected value. To make is clear, how about introducing a new local variable e.g. `tmp_errno`, store errno into `tmp_errno` in line 190, and print the value here?

mwarning: I've added `connect_errno`. Let me know if this is ok.

mwarning:
Thanks for merging!

But if p == NULL, and it connect was not the one that failed (rare - but not impossible), then we have an error message with an "no error" message (connect_errno == 0).

So this might be better:
```
if (p == NULL) {
    if (connect_errno == 0) {
       dlt_log(LOG_ERR, "ERROR: failed to connect.\n");
    } else {
       dlt_vlog(LOG_ERR, "ERROR: failed to connect! %s\n", strerror(connect_errno));
    }
}
````

But it is ok for me as it is. I won't complain. :-)

PR #267, opened 2020-11-10
Android adapter: Init missing buffer + message length check

tobohr:
Some improvements for android adapter

- Earlier the crash buffer wasn't initalized,
claimed to be binary which isn't true.
https://cs.android.com/android/platform/superproject/+/master:system/logging/logcat/logcat.cpp;drc=5837d33d1cfecce0632e385c8ede89c02685238f;l=194
- Also no need register buffers which isnt used.
- Adds message length check to handle malformed message.

### DIFF
--- a/src/android/dlt-logd-converter.cpp
+++ b/src/android/dlt-logd-converter.cpp
@@ -167,9 +167,15 @@ static int logd_parser_loop(struct logger_list *logger_list)
         DltLogLevelType log_level;
         log_level = get_log_level_from_log_msg(&log_msg);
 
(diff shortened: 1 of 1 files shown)


ssugiura:
@tobohr Can you add signed-off commit message to your first commit like you've done in https://github.com/GENIVI/dlt-daemon/pull/267/commits/d4d5b5fc9c236cd9a86a969e3a68d94f38f5ebed? Thanks.

tobohr:
@ssugiura yes sorry apparently missed on that commit, but now its in.

PR #269, opened 2020-11-12
Issue 265

schrewe:
This changes should fix issue 265. https://github.com/GENIVI/dlt-daemon/issues/265
By adding a length check the write buffer overflow is prevented.
Additionally the way the ring buffer is increased is changed because the case of increasing one time is not sufficient was missed.

### DIFF
--- a/src/shared/dlt_common.c
+++ b/src/shared/dlt_common.c
@@ -2601,7 +2601,7 @@ int dlt_buffer_push3(DltBuffer *buf,
         free_size = buf->size - write + read;
 
     /* check size */
-    if (free_size < (int)(sizeof(DltBufferBlockHead) + size1 + size2 + size3)) {
+    while (free_size < (int)(sizeof(DltBufferBlockHead) + size1 + size2 + size3)) {
         /* try to increase size if possible */
         if (dlt_buffer_increase_size(buf))
             /* increase size is not possible */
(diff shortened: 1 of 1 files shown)


ssugiura:
@schrewe We are currently looking into your changes deeply. In parallel, can you add signed-off to your commit message like in other commits? Thanks.

schrewe:
@ssugiura of course, is it ok this way?

ssugiura:
@schrewe Thanks. Though CodeQL is complaining about some lines. And your devel branch seems to be way behind current master branch. Could you rebase it to latest so that we're straight forward to go once our evaluation is done?

src/shared/dlt_common.c
```diff
@@ -2529,7 +2534,7 @@ int dlt_buffer_push3(DltBuffer *buf,
         free_size = buf->size - write + read;
 
     /* check size */
-    if (free_size < (int)(sizeof(DltBufferBlockHead) + size1 + size2 + size3)) {
+    while (free_size < (int)(sizeof(DltBufferBlockHead) + size1 + size2 + size3)) {
```
Kevin-Luong, line 2537: Thank you @schrewe . I think we should also update the free_size variable inside the while loop (or use do-while) to prevent the infinitive loop for this case :)

schrewe:
@Kevin-Luong thanks for the hint, now the free_size is updated in the while loop.
@ssugiura I rebased on master.

ssugiura:
@schrewe Indention is not correct for your new lines, could you update the lines? Other than that, the implementation itself seems reasonable.

schrewe:
@ssugiura Ok, cool! I fixed the identation issue.

PR #271, opened 2020-11-19
Use dlt_defaults for dlt-logd-converter in Android.bp

martin-ejdestig:


ssugiura:
@martin-ejdestig Can you give brief overview of what you want to achieve with proposed change?

martin-ejdestig:
Just think it is an oversight that the converter is the only target in Android.bp that does not use dlt_defaults.

(In our internal branch we change "dlt_defaults" (we have internal defaults for all external components, nothing that makes sense to upstream) and at first we did not understand why the converter was not affected. That is how this was noticed.)

PR #284, opened 2021-01-03
dlt_client_main_loop running in an infinite loop restricts graceful e…

sreeharshar84:
…xit of

DLT Client code

Issue:
------
dlt_client_main_loop currently uses an infinite loop ("while (1)"). This
creates problems when a DLT client is running in a thread as part of a larger
application. Graceful exit (for example: during object destruction) is not
possible because a thread is running dlt_client_main_loop in the background.

It is also not possible to exit the client, if we want it to fetch only a
pre-decided number of messages.

Solution:
---------
Allow user to define a callback function to check whether the next message
should be fetched.

dlt-test-client.c has a new option to fetch a pre-decided number of messages,
after which the client will exit.

### DIFF
--- a/include/dlt/dlt_client.h
+++ b/include/dlt/dlt_client.h
@@ -75,6 +75,7 @@
 
 #   include "dlt_types.h"
 #   include "dlt_common.h"
+#include <stdbool.h>
 
 typedef enum
 {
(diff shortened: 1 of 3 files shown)


ssugiura:
@sreeharshar84 Could you add signed-off to your commit message? The change itself looks good to me.

sreeharshar84:
> @sreeharshar84 Could you add signed-off to your commit message? The change itself looks good to me.

@ssugiura did you get a chance to see the updated commit message. 

I did a "git commit --ammend" and a force push. I wasn't sure if this was the process you followed. 

In case this is not the process, I can submit a fresh Pull Request.

PR #285, opened 2021-01-15
Make dlt pipe readable only by owner

idleroamer:
Security-wise the dlt pip(FIFO) should be exclusively only readable by user owner not the whole group,
in order to prevent eavesdropping by clients


ssugiura:
@idleroamer The change looks good to me. Can you please add signed-off to your commit message? I can merge right away. Thank you!

idleroamer:
@ssugiura Thanks for your support.

PR #290, opened 2021-02-10
Fix include path in *.pc file

mawillers:
Previously, an application using this library's pkg-config setting for
include directory would get e.g. -I/usr/local/dlt/include/dlt
With that include dir, the application needs to write its #include
statements without any prefix, i.e. "#include "dlt_user.h"

This is considered bad practice for header files of an elaborate
library consisting of multiple header files.
For such libraries, one should write #include statements like
this: #include "dlt/dlt_user.h", i.e. prefix it with a directory.

This commit achieves that. The include directory returned by the
above example is now -I/usr/local/dlt/include

### DIFF
--- a/automotive-dlt.pc.in
+++ b/automotive-dlt.pc.in
@@ -23,5 +23,5 @@ Description: Diagnostic Log and Trace
 Version: @PROJECT_VERSION@
 Requires:
 Libs: -L${libdir} -ldlt -lrt -lpthread @ZLIB_LIBRARY@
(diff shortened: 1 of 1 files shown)


ssugiura:
@mawillers I believe backward compatibility would break with this change and this impacts a lot of applications which is already using DLT.

mawillers:
@ssugiura alright, we can drop this PR, but what about #289?

ssugiura:
@mawillers Sorry for late update. I've merged https://github.com/GENIVI/dlt-daemon/pull/289, so I will drop this PR. Thanks for your contribution!

PR #291, opened 2021-02-10
Better formatting of RAWD

mawillers:
Previously, the bytes of a RAWD argument were be printed onto the console
as space-separated hex values, making it difficult or impossible to see
where the RAWD argument ends and the next argument starts. For instance,
with code like this:

    dlt_user_log_write_raw(&ctxData, "data", 4);
    dlt_user_log_write_uint8(&ctxData, 53);

the console output is shown as:

    [64 61 54 61 53]

and there is no indication that there are two arguments, one RAWD and
one UINT8.

This patch changes for console formatting of RAW data to separate by
apostrophs, making the above message appear as:

    [64'61'54'61 53]

This makes it immediately obvious that there is a RAWD argument, and
still keeps the separators quite unobtrusive.

### DIFF
--- a/src/shared/dlt_common.c
+++ b/src/shared/dlt_common.c
@@ -3881,7 +3881,7 @@ DltReturnValue dlt_message_argument_print(DltMessage *msg,
         if ((*datalength) < length)
             return DLT_RETURN_ERROR;
 
-        dlt_print_hex_string(value_text, (int) textlength, *ptr, length);
+        dlt_print_hex_string_delim(value_text, (int) textlength, *ptr, length, '\'');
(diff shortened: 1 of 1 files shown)


ssugiura:
@mawillers This change is applied to not only RAW output but general hex message printing e.g. dlt-receive with hex and/or string output, and I would like to keep their format as is. What I can propose here is to introduce function like `dlt_print_hex_string_w_delimiter()` and replace with apostrophes only for RAW output.

mawillers:
@ssugiura, I have implemented your suggestion. Please re-check.

PR #293, opened 2021-02-19
dlt-receive: Add option to configure port of dlt receive

alexmohr:
This allows the dlt-receive command line tool to be used if the dlt-daemon is not listening on the default port.

### DIFF
--- a/doc/dlt-receive.1.md
+++ b/doc/dlt-receive.1.md
@@ -6,7 +6,7 @@
 
 # SYNOPSIS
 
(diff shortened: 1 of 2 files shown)


ssugiura:
@alexmohr Thanks for contribution. The change itself looks good to me. Two minor requests:
- Update the document as well https://github.com/GENIVI/dlt-daemon/blob/master/doc/dlt-receive.1.md
- Add signed-off to your commit message

alexmohr:
@ssugiura Thanks for your feedback, I've added the requested changes

PR #301, opened 2021-04-06
Find system-provided external gtest in case a system version is avail…

LocutusOfBorg:
…able, without need to recompile the bundled version everytime

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,7 +240,16 @@ add_subdirectory(src)
 add_subdirectory(include/dlt)
 add_subdirectory(testscripts)
 if (WITH_DLT_UNIT_TESTS)
-    add_subdirectory( gtest-1.7.0 )
+    find_path(GTEST_ROOT src/gtest.cc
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -240,7 +240,16 @@ add_subdirectory(src)
 add_subdirectory(include/dlt)
 add_subdirectory(testscripts)
 if (WITH_DLT_UNIT_TESTS)
-    add_subdirectory( gtest-1.7.0 )
+    find_path(GTEST_ROOT src/gtest.cc
```
ssugiura, line 243: I haven't fully understood the sequence here. What is `GTEST_ROOT` needed for?

LocutusOfBorg: I think this helps the find_package GTest.cmake to correctly find and build gtest on all systems!
```
The following cache variables may also be set:

``GTEST_ROOT``
  The root directory of the Google Test installation (may also be
  set as an environment variable)
```

PR #311, opened 2021-06-09
Genivi upstream minor version 2.18.7

tdinhcong:
Hi @ssugiura -san and @thanhbnq , 
This is release for version v2.18.7
Please kindly have a review and take other step. 

Thank you.


tdinhcong:
Hi @MilanRusko, 
Thank you for your comment. I would like to update some points:

1. The first patch in your comments seems not have any difference compared with the code change in the commit. And if you need to update this, please comment directly **block of line** you want, it will be more easier to tracking.

2. Currently, in dlt_buffer_write_block(), there will be **two** cases can be happen: 
* Write single block of data to buf->mem (a buffer for shared memory) with condition below:
> if (( (unsigned int) (*write ) + size) <= buf->size)

* Write two block of data to buf->mem if condition above is wrong. In this case, ringbuffer size (buf->size) is not enough to store the data, so it will write a part of data to buf->mem to fill remain space in ringbuffer first, and then write the data left at the beginning position.

After that, buf->size and write position will be updated later. So I think we don't need to have change here.

Please let me know if you have any concerns. Thank you.

MilanRusko:
Hi,

sorry there was a mistake in 1), I fixed it and write comments.

I insist, there is a mistake in the code. Just take in mind this example:
1. call dlt_buffer_write_block() with size=buf->size (when *write==0)
   Result:  *write = buf->size // it is wrong, because write pointer can be
from 0 to buf->size-1 (we cannot write to buf->size position)
   But before this bug was automatically fixed when
dlt_buffer_write_block() was called second time in the second if branch
2. then call  dlt_buffer_write_block() with size>0 (when *write==buf->size)
   first if condition fails if (( (unsigned int) (*write ) + size) <=
buf->size) == false
   second if condition fails if(buf->size > (unsigned int) (*write)) ==
false
   So no data are written into the buffer!!
   And this happens any time when  dlt_buffer_write_block() is called!!
   The circular buffer is blocked and dysfunctional!!


ut 15. 6. 2021 o 5:56 tdinhcong ***@***.***> napísal(a):

tdinhcong:
Hi @MilanRusko, 
I think I get your point here. You are right that when we call **dlt_buffer_write_block()** with size = buf->size  and *write==0, this can lead to the issue. 
But the condition ```size = buf->size``` and ``` *write = 0``` is **never** happen, as my observasion.
In this line [2683](https://github.com/GENIVI/dlt-daemon/blob/3bb7eb24092ff771ccfd1119f8c1088f3f78f981/src/shared/dlt_common.c#L2683), we always write the header data first, and after we will write other data.
* With the first message will be written to ringbuffer, this condition : 
``` header size < (buf -> size)```
is always right, to make sure that at least header of this message can be written. So condition ```*write = 0``` is right, but  ```size = buf->size ``` is wrong.
* After that, write position will always be updated, and can not reach ```0``` again. Buffer in dlt will notify that buffer is full and do not receive and messages, this make write position can not update to ```0```.

It is very glad that you can suggest the specific config and condition can leads to this issue. In this case, I can verify clearly.
Thank you.

**Comment by MilanRusko** [comment]
OK, the condition   size = buf->size and *write = 0 is *never* happen
But *write can be set to buf->size by other way, pls check the comment
https://github.com/tdinhcong/dlt-daemon/commit/6341d9f9e2905ed6d18b7b8ed00e5de48ebd3326#commitcomment-52226320

Milan

ut 15. 6. 2021 o 13:48 tdinhcong ***@***.***> napísal(a):

File: src/shared/dlt_common.c [hunk 1]
```diff
@@ -2431,7 +2431,10 @@ void dlt_buffer_write_block(DltBuffer *buf, int *write, const unsigned char *dat
                 *write += (int) size;
             }
             else {
-                if(buf->size > (unsigned int) (*write)) {
+                /* when (*write) = buf->size, write only the second block
+                * and update write position correspondingly.
+                */
+                if((unsigned int) (*write) <= buf->size) {
```
tdinhcong, line 2437: Hi @MilanRusko , 
I have updated this commit for correcting write/read position. Can you take a look on it and give feedback? 

Thank you in advance.

ssugiura:
@MilanRusko Do you have further feedback to this? Otherwise it seems to be ready to be merged, Thanks.

PR #319, opened 2021-07-07
Extend include path in *.pc file

### LINKED ISSUE #318: CMake, pkg-config: Support #include <dlt.h> and #include <dlt/dlt.h>
`doc/dlt_for_developers.md` states:

> Please note that the #include statement depends on the means by which you are incorporating the DLT library into your project. The <dlt/dlt.h> form (i.e. with a directory prefix) seen here is necessary when you are using the CMake Config file (see below). If you are using pkg-config instead, then this #include statement needs to refer to only <dlt.h>, due to the way the pkg-config module exports the include directory.

This paragraph completely ignores that code reuse is a common practice. For example, the use of a DLT based code with pkg-config that was previously developed using CMake. This is not easily possible. One has to change the include statement to make the code compile.

This was added in #289. The include statement was previously discussed in #205.

Please consider the exported include path, names and relative locations of the public header files to be a part of the DLT API. The API should not depend on whether one uses pkg-config or CMake since this creates unnecessary integration efforts and hampers code reuse.

Due to this I suggest the following:

* The documentation should state that `#include <dlt/dlt.h>` should be used.
* For backwards compatiblity `#include <dlt.h>` also works, but it's use is deprecated.
* Export both include paths in the pkg-config and CMake config file.

mawillers:
Previously, an application using this library's pkg-config setting for
include directory would only get e.g. -I/usr/local/dlt/include/dlt
With that include dir, the application needs to write its #include
statements without any prefix, i.e. "#include "dlt_user.h"

This is considered bad practice for header files of an elaborate
library consisting of multiple header files.
For such libraries, one should write #include statements like
this: #include "dlt/dlt_user.h", i.e. prefix it with a directory.

This commit achieves that by letting the pkg-config module also return
the upper-level directory in addition to the previous one.
For instance, the include directive returned by the above example
would now be "-I/usr/local/dlt/include -I/usr/local/dlt".

Closes #318

### DIFF
--- a/automotive-dlt.pc.in
+++ b/automotive-dlt.pc.in
@@ -23,5 +23,5 @@ Description: Diagnostic Log and Trace
 Version: @PROJECT_VERSION@
 Requires:
 Libs: -L${libdir} -ldlt -lrt -lpthread @ZLIB_LIBRARY@
-Cflags: -I${includedir}/dlt -DDLT_@PROJECT_VERSION_MAJOR@_@PROJECT_VERSION_MINOR@
+Cflags: -I${includedir}/dlt -I${includedir} -DDLT_@PROJECT_VERSION_MAJOR@_@PROJECT_VERSION_MINOR@
 
--- a/doc/dlt_for_developers.md
+++ b/doc/dlt_for_developers.md
@@ -28,9 +28,9 @@ this document.
 Please note that the #include statement depends on the means by which you are
 incorporating the DLT library into your project. The `<dlt/dlt.h>` form (i.e.
 with a directory prefix) seen here is necessary when you are using the CMake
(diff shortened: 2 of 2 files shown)


sebastianlipponer:
Thanks!

CMake should also export an additional include path to support legacy includes.
https://github.com/GENIVI/dlt-daemon/blob/57d981f9d765286a85e682f7bee0769d7349e7aa/src/lib/CMakeLists.txt#L53
`$<INSTALL_INTERFACE:include/dlt>` is missing.

Moreover, `doc/dlt_for_developers.md` must also be changed to state that legacy includes like `#include <dlt.h>` are supported for backwards compatibility, but should not be used anymore.

PR #325, opened 2021-08-27
dlt_user_shared: Add timeout to writev

alexmohr:
This timeout is necessary to prevent blocking writev indefinitely.
Without the timeout dlt-daemon, may block indefinitely when an app id is re-used
very frequently.
In that case dlt-daemon won't accept new connections and further communication
in any way is not possible anymore.

The setup of timeouts for sockets in `dlt_daemon_process_client_connect`
is not enough. 
Changed calls still can block for ever. 
It might be better to use timeoutOnSend but this would require 
more refactoring to pass the timeout to the log calls. 

Alexander Mohr, [alexander.m.mohr@daimler.com](mailto:alexander.m.mohr@daimler.com), Daimler TSS GmbH, [imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)

Signed-off-by: Alexander Mohr <alexander.m.mohr@daimler.com>>

### DIFF
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -837,12 +837,17 @@ static int dlt_mkdir_recursive(const char *dir)
     for (p = tmp + 1; ((*p) && (ret == 0)) || ((ret == -1 && errno == EEXIST) && (p != end)); p++)
         if (*p == '/') {
             *p = 0;
-            ret = mkdir(tmp,
-            #ifdef DLT_DAEMON_USE_FIFO_IPC
-                        S_IRWXU);
-            #else
-                        S_IRUSR | S_IWUSR | S_IXUSR | S_IRGRP | S_IWGRP | S_IROTH  | S_IWOTH /*S_IRWXU*/);
-            #endif
+
+            struct stat s;
+            if (stat(tmp, &s) == 0 && !S_ISDIR(s.st_mode)) {
+                ret = mkdir(tmp,
+                #ifdef DLT_DAEMON_USE_FIFO_IPC
+                                S_IRWXU);
+                #else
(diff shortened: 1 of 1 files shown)


ssugiura:
@alexmohr Can you give more detailed scenario why the socket could be blocked? Why Application IDs could be re-used frequently?

alexmohr:
> @alexmohr Can you give more detailed scenario why the socket could be blocked? Why Application IDs could be re-used frequently?

We where not really able to find out _why_ the sockets are getting blocked. 
The dlt-daemon stopped communication in all regards. 
After taking a look with a debugger we saw the it's blocked in `writev` every time the problem occurs.

The re-using of the application IDs is due to the fact the same application is started multiple times while it's already running. 
Probably there is a way to fix the application that's causing the problem too.
But no application no matter how bad it's programed should be able to DOS the dlt-daemon.

The only way we could reproduce the problem is if dlt-daemon is started via systemd on bootup. Once it's restarted or started manually the problem does not show up anymore. 
The system with this issue also does not have any firewall or strange mounting rules that could cause the problem.

I've tested the suggested change for 18h and could not reproduce the blocking of the sockets anymore. Before the fix it happened on every boot after ~30m-1h depending on how often an application id was re-used.

The system where we've seen this bug is still using `v2.18.5` but the modified code hasn't change between `v2.18.5` and `master`

To reproduce the issue we applied the patch attached to this issue and ran the following command on the target after bootup.
```
watch "/tmp/dlt-example-user -n 50 -d 50 localhost  & sleep 1 &&  /tmp/dlt-example-user -n 50 -d 50 localhost"
```

[0001-dlt-example-user-Remove-all-unregister-calls.zip](https://github.com/GENIVI/dlt-daemon/files/7074637/0001-dlt-example-user-Remove-all-unregister-calls.zip)

alexmohr:
Closing this PR as I've overwrote the master branch and there was no interaction in the last 2 Months

PR #326, opened 2021-08-28
dlt_user: Replace prctl() with pthread_setname_np()

sebastianlipponer:
Use POSIX thread API to rename the housekeeper and segmented threads. The currently used prctl() function is Linux specific and in general not available on other platforms (e.g. QNX).


src/lib/dlt_user.c
```diff
@@ -4708,6 +4698,11 @@ int dlt_start_threads()
         dlt_log(LOG_CRIT, "Can't create housekeeper thread!\n");
         return -1;
     }
+    if (pthread_setname_np(dlt_housekeeperthread_handle, "dlt_housekeeper"))
```
ssugiura, line 4701: Any reason you moved from `dlt_user_housekeeperthread_function`?

src/lib/dlt_user.c
```diff
@@ -4708,6 +4698,11 @@ int dlt_start_threads()
...
+    {
+        dlt_log(LOG_CRIT, "Can't rename housekeeper thread!\n");
+        return -1;
```
ssugiura, line 4704: I wonder if it should return error in case the thread name could not be renamed

sebastianlipponer: An error may occur if the name does not fit to what the OS can handle. This is probably mostly an issue when running the code the first time on another platform. I think it's nice to handle this properly.  On the other hand, you are right, it not a critical error though, the code still works without renaming.

src/lib/dlt_user.c
sebastianlipponer: I do not see how to handle an error in the thread functions themselves. Also I think it's a bit nicer to do the renaming together with the thread creation.

src/lib/dlt_user.c
ssugiura: I also see that some OS, even some Linux system, does not support `pthread_setname_np()`, so we need to consider that case too, otherwise it will fail to compile.
We internally have a solution to check existence of the call, let me raise a separate PR for it and you can change on top of that. How does that sound? :)

ssugiura: FYI: https://github.com/GENIVI/dlt-daemon/pull/329

sebastianlipponer: Thanks for pointing this out. It looks like `pthread_setname_np()` was previously already in use. I have updated the change accordingly.

PR #330, opened 2021-09-14
Update dlt-daemon.c

kundatipradeep:
As init_p2 does the offline trace initialization, load configuration should be called prior, as LoggingMode is set through run-timeconfiguration file.  With original logic it results in offline trace write failed, it is not valid trace directory errors.

### DIFF
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -970,6 +970,12 @@ int main(int argc, char *argv[])
         dlt_log(LOG_ERR, "Could not load runtime config\n");
         return -1;
     }
+ 
+    /*
+     * Load dlt-runtime.cfg if available.
+     * This must be loaded before offline setup
+     */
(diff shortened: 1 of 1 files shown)


src/daemon/dlt-daemon.c
```diff
@@ -979,13 +985,7 @@ int main(int argc, char *argv[])
...
     /* --- Daemon init phase 2 end --- */
 
-    /*
-     * Load dlt-runtime.cfg if available.
-     * This must be loaded before offline setup
-     */
-    dlt_daemon_configuration_load(&daemon, daemon.runtime_configuration, daemon_local.flags.vflag);
-
-    if (daemon_local.flags.offlineLogstorageDirPath[0])
+   if (daemon_local.flags.offlineLogstorageDirPath[0])
```
ssugiura, line 988: Unexpected change here

PR #332, opened 2021-09-18
Make the legacy include path a CMake option

mawillers:
Keep the generated CMake config file "clean" in that only `<prefix>/include/` is being added to the compiler's header search path by default. If the legacy include path, i.e. `<prefix>/include/dlt/`, is also to be added, then use the CMake option `-DWITH_LEGACY_INCLUDE_PATH=On`.

See also #327, #319, #318

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,8 @@ option(WITH_DLT_KPI           "Set to ON to build src/kpi binaries"             
 option(WITH_DLT_FATAL_LOG_TRAP "Set to ON to enable DLT_LOG_FATAL trap(trigger segv inside dlt-user library)"    OFF)
 option(WITH_UDP_CONNECTION     "Set to ON to enable dlt UDP multicast SUPPORT"                                   OFF)
 option(WITH_LIB_SHORT_VERSION "Set to ON to build library with only major number in version"                    OFF)
+option(WITH_LEGACY_INCLUDE_PATH "Set to ON to add <prefix>/dlt to include paths for the CMake config file, in addition to only <prefix>"
+   OFF)
 option(WITH_EXTENDED_FILTERING "Set to OFF to build without extended filtering. Using json filter files is only supported for Linux based system with json-c and QNX." OFF)
 
 option(WITH_DLT_DAEMON_VSOCK_IPC "Set to ON to enable VSOCK support in daemon"                                   OFF)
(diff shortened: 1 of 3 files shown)


sebastianlipponer:
The CMake config file is already clean. This was discussed in thoroughly in #318. Your definition of clean is obviously in the interest of your employer.

Currently, both include paths are exported and users can rely on that. For example, a user can install a dlt-daemon version available in a Linux distribution and include statements work as expected regardless of whether pkg-config or CMake config files are used.

With this option this is not the case anymore since we don't know what a package maintainer will configure. I don't see _any_ added value. This change basically reverts #327 and reintroduces the issue #318 again.

mawillers:
I see that your definition of clean is apparently in only your interest, as you seem to insist on using a completely new feature (a CMake config file) in a way (compatible to a broken codebase) that inflicts pain on everyone else.
If you cannot convert your codebase to use the proper way of writing #include statements, then don't use the CMake config file and stick with the pkg-config one instead.

sebastianlipponer:
_The opposite is true_: Consistent pkg-config and CMake config files makes everyone else's life easier except yours. Proof is that you are the only one talking about 'pain' while no one complained about clashing headers for years with the existing pkg-config file.

@ssugiura Please do not merge this change.

mawillers:
No, broken CMake config files do _not_ make life easier for anyone.

You still haven't explained to us why you don't want to keep using the pkg-config file? That works exactly as you'd like it to.

sebastianlipponer:
I have raised #318 with a full explanation. This is about backwards compatibility!

The CMake config file is backwards compatible and consistent with the pkg-config file. That's why #327 was merged in the first place. Users are now able migrate to the new include statement step by step and can choose to adopt the CMake config file without being forced to change their code at the same time.

You should give a proper explanation of the problem without using fuzzy terms like 'broken', 'pain', 'not clean' and 'cluttered'. Moreover, your argument should not be based on the integration of a broken in-house code that no one cares about.

mawillers:
> This is about backwards compatibility!

No, it is not, please do not spread misinformation here.
Until a few weeks ago, there was no CMake config file at all. Hence, there cannot be any backward incompatibility with it.

> You should give a proper explanation of the problem without using fuzzy terms like 'broken', 'pain', 'not clean' and 'cluttered'.

A broken build system is one where I get undesired include paths injected into my include search path. It is "painful" to work with, because I have to use inside knowledge about the CMake config file in order to work around that, and such exploitation of internal implementation details are, of course, fragile and tend to break with new releases of that file.
A "clean" build system is one where I only get exactly those paths that are needed and desired, and nothing else, and where I can use all the libraries in the world together in a single application without experiencing any naming conflicts.

You, however, still have not explained to us why you do not want to stick with the pkg-config setup. That is still working as before and was, in fact, the only choice anyway until a few weeks ago.

> Moreover, your argument should not be based on the integration of a broken in-house code that no one cares about.

A library that provides a header called "dlt.h" is not broken. It is entirely reasonable to do so, because the name/acronym "dlt" is not reserved for use by only Genivi's libdlt. The world does not revolve around Genivi DLT, nor around automotive software, and also not about Debian, for that matter.
Case in point: a 10 seconds Google search showed two other Open Source header files called "dlt.h":
https://ftp.netbsd.org/pub/NetBSD/NetBSD-current/src/sys/net/dlt.h
https://github.com/the-tcpdump-group/libpcap/blob/master/pcap/dlt.h
You really should avoid being so narrow-minded, and have a glimpse beyond the boundaries of your own scope. You don't care about any other software project providing a similarly-named header file? - Fine, but there are many people who do care, and they will not be pleased by such high-handedness that you are showing right now.

The current *.pc and CMake config file _are_ broken, however, because they actively prevent me from using any library that also provides a header called "dlt.h"


Again: Why can't you stick with pkg-config? That works exactly as you want it to.

ssugiura:
@mawillers @sebastianlipponer 
Thanks a lot for your proposal and discussion on this topic.
I also know other projects are seeing similar behavior with the change of include path, though it is indeed true that cmake config file was _officially_ newly introduced lately.
So I'd like to keep as default to keep current code to include also include/dlt.
Would it be possible to set this new cmake option to ON as default, and set to OFF at @mawillers 's setup?

PR #333, opened 2021-09-27
dlt-daemon: create sockets using "android way"

sebastienraillet:
# Summary
The goal of this PR is to use the android way to create sockets inside DLT daemon for android system to allow a better configuration of SELinux labels / contexts on them.

# Description
* Before this PR, dlt-daemon sockets were created inside /data/local/tmp. This works but have many drawbacks:
    - /data/local/tmp isn't always a tmpfs depending of the android system you have. Means sockets are potentially created on a  filesystem which binds to a real device
    - as the sockets are created by the dlt-daemon itself, this prevent the usage of specific SELinux labels / contexts as they will inherit the label / context from its parent location (e.g the ones from /data/local/tmp). This prevent a fine control of the SELinux label / context that you would like to apply on them

* This commit adapts the dlt-daemon in such way: 
    - application and control sockets are now created inside /dev/socket which is the standard path for sockets on android
    - these sockets are now created by init (see dlt-daemon.rc) and their fds are recovered by dlt-daemon through a specific android API (dlt_daemon_unix_android_get_socket). If the fds can't be recovered, we fallback to the previous mechanism by creating by ourself the sockets (even if this will prevent SELinux label / context on this socket)
    - all these modifications have been put under compilation flag for android

Signed-off-by: Sebastien RAILLET <sebastien.raillet@marelli.com>

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -29,7 +29,7 @@ cc_defaults {
         "-DDLT_DAEMON_USE_UNIX_SOCKET_IPC",
         "-DDLT_LIB_USE_UNIX_SOCKET_IPC",
         "-DCONFIGURATION_FILES_DIR=\"/vendor/etc\"",
-        "-DDLT_USER_IPC_PATH=\"/data/local/tmp\"",
+        "-DDLT_USER_IPC_PATH=\"/dev/socket\"",
     ] + [
         "-Wno-unused-parameter",
         "-W",
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -1436,12 +1436,111 @@ static int dlt_daemon_init_vsock(DltDaemonLocal *daemon_local)
 }
 #endif
 
+#ifdef DLT_DAEMON_USE_UNIX_SOCKET_IPC
+static DltReturnValue dlt_daemon_init_app_socket(DltDaemonLocal *daemon_local)
+{
+    /* socket access permission set to srw-rw-rw- (666) */
+    int mask = S_IXUSR | S_IXGRP | S_IXOTH;
+    DltReturnValue ret = DLT_RETURN_OK;
+    int fd = -1;
+
+    if (daemon_local == NULL) {
+        dlt_vlog(LOG_ERR, "%s: Invalid function parameters\n", __func__);
+        return DLT_RETURN_ERROR;
+    }
+
+#ifdef ANDROID
+    /* on android if we want to use security contexts on Unix sockets,
+     * they should be created by init (see dlt-daemon.rc in src/daemon)
+     * and recovered through the below API */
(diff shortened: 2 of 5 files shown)


Android.bp
```diff
@@ -29,7 +29,7 @@ cc_defaults {
         "-DDLT_DAEMON_USE_UNIX_SOCKET_IPC",
         "-DDLT_LIB_USE_UNIX_SOCKET_IPC",
         "-DCONFIGURATION_FILES_DIR=\"/vendor/etc\"",
-        "-DDLT_USER_IPC_PATH=\"/data/local/tmp\"",
+        "-DDLT_USER_IPC_PATH=\"/dev/socket\"",
```
thanhbnq, line 32: Thank you @sebastienraillet 
What do you think about "/data/vendor"?

sebastienraillet: Hello @thanhbnq,
Of my understanding, _/data/vendor_ is more about storing specific data for native applications (e.g internal database, binary files, etc.).

I've use _/dev/socket_ as it is the location where init daemon creates socket declared inside the .rc file (see [here](https://cs.android.com/android/platform/superproject/+/master:system/core/init/util.cpp;drc=3b3d87de79ab993eb1dc03286deb6f6c0e8c7c68;l=89) and [here](https://cs.android.com/android/platform/superproject/+/master:device/generic/vulkan-cereal/fake-android-guest/cutils/sockets.h;drc=9958a2d1d95a4c9ad1578ffe6cc0b78e7368e22b;l=46))

Now for the sake of curiosity 😉, I've tried to move the sockets inside _/data/vendor_ to understand what are the drawbacks in this case. You can find below a summary : 
## sockets inside /dev/socket
* sockets are created and setup (permissions, mode, etc.) by init process
* recovered by dlt-daemon through _**android_get_control_socket()**_ API
* allow to specify one SELinux label for each sockets (this is done by init when sockets are created) 
## sockets inside /data/vendor
* a specific folder inside /data/vendor is created by init (e.g dlt folder)
* sockets are created inside this folder by dlt-daemon as usual
* dlt-daemon sockets will inherit the SELinux label of the folder. Means we can't put different labels on each sockets.

Of my point of view, in term of security, and based on the principale of least privilege, it is better to be able to apply distinct SELinux label for each socket. Like that, you can split easily application having access to dlt app socket and dlt control socket.
By checking also the aosp documentation [here](https://source.android.com/security/selinux/compatibility#file-ownership), they are talking about _/dev/vendor/socket_. I'm not able to find this path neither on Android 10 nor in Android 11. Knowing that this documentation has been updated on 22 septembre 2021, maybe this is something planned for Android 12 🤞 . Means that vendor applications will potentially have a specific location for their sockets in the future.

But maybe you had in mind some corner cases I've not in mind on my side where _/dev/socket_ doesn't sounds a good location 🙂 ?

PR #345, opened 2021-10-22
New data channel based on QNX Messaging

svstanko:
- Add a new DLT-IPC "QNX_MESSAGE" based on the QNX messaging IPC
in addition to the two existing DLT-IPC: FIFO and UNIX_SOCKET.

The problem:
Message passing is the primary form of IPC in the QNX Neutrino
RTOS. POSIX IPC (TCP/IP sockets and pipes) are built over QNX
native message passing and sometimes it can cause 2 copies of
data, more context switches.

The solution:
Using of Resource Manager helps to simplify client-server
communication and switch DLT to QNX Neutrino Messaging instead
Unix sockets with minimum efforts without changing in DLT internal
logic.

The conclusions:
The main benefit of replacing Unix sockets with QNX Neutrino
Messaging is significant unloading of the networking manager
(io-pkt-v6-hc):
- CPU Usage ~ 9 times better;
- Inter CPU Communication (Cross CPU Messages) ~68 times better;
- CPU Migration ~ 2 times better.

DLT daemon utilizes more CPU recourses (instead of the network
driver), but this has not degraded overall performance.

Noticeable improvement in CPU usage has been achieved.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ option(WITH_DLT_TESTS         "Set to ON to build src/test binaries"            
 option(WITH_DLT_UNIT_TESTS    "Set to ON to build gtest framework and tests/binaries"                            OFF)
 option(WITH_DLT_QNX_SYSTEM    "Set to ON to build QNX system binary dlt-qnx-system"                              OFF)
 
-set(DLT_IPC "FIFO" CACHE STRING "UNIX_SOCKET,FIFO")
+set(DLT_IPC "FIFO" CACHE STRING "UNIX_SOCKET,FIFO,QNX_MESSAGE")
 set(DLT_USER "genivi" CACHE STRING "Set user for process not run as root")
 
 option(WITH_DLT_PKGCONFIG     "Set to ON to generate pkgconfig .pc files"                                        ON)
@@ -121,7 +121,7 @@ include_directories(
 
 add_definitions(-D_GNU_SOURCE)
 
-if(NOT DLT_IPC STREQUAL "UNIX_SOCKET" AND NOT DLT_IPC STREQUAL "FIFO")
+if(NOT DLT_IPC STREQUAL "UNIX_SOCKET" AND NOT DLT_IPC STREQUAL "FIFO" AND NOT DLT_IPC STREQUAL "QNX_MESSAGE")
     message(FATAL_ERROR "${DLT_IPC} is not a valid value for DLT_IPC")
 endif()
 add_definitions(-DDLT_DAEMON_USE_${DLT_IPC}_IPC)
--- a/include/dlt/dlt_common.h
+++ b/include/dlt/dlt_common.h
@@ -111,6 +111,16 @@
 #      define DLT_PACKED
 #   endif
 
+#   ifdef DLT_DAEMON_USE_QNX_MESSAGE_IPC
+#      ifdef __QNX__
+#          include <sys/iofunc.h>
+#          include <sys/resmgr.h>
+#          include <sys/dispatch.h>
+#      else
+#          error DLT_IPC="QNX_MESSAGE" can only be built for QNX
+#      endif /* __QNX__ */
+#   endif /* DLT_DAEMON_USE_QNX_MESSAGE_IPC */
+
 /*
  * Macros to swap the byte order.
  */
@@ -410,6 +420,10 @@ typedef enum
     DLT_RECEIVE_SOCKET,
     DLT_RECEIVE_UDP_SOCKET,
     DLT_RECEIVE_FD
+#ifdef DLT_DAEMON_USE_QNX_MESSAGE_IPC
+    DLT_RECEIVE_MSG,
+    DLT_CLIENT_RECEIVE_MSG,
+#endif /* DLT_DAEMON_USE_QNX_MESSAGE_IPC */
 } DltReceiverType;
 
 /**
@@ -784,6 +798,14 @@ typedef struct
     DltReceiverType type;     /**< type of connection handle */
     int32_t buffersize;       /**< size of receiver buffer */
     struct sockaddr_in addr;  /**< socket address information */
+#ifdef DLT_DAEMON_USE_QNX_MESSAGE_IPC
+    resmgr_context_t *ctp;    /**< Context information that's passed between resource-manager functions */
+    int offset;               /**< offset for resmgr_msgread() */
+    int nbytes;               /**< received bytes  for resmgr_msgread() */
+    void *daemon;             /**< pointer to DltDaemon */
+    void *daemon_local;       /**< pointer to DltDaemonLocal */
+    name_attach_t *attach;
+#endif /* DLT_DAEMON_USE_QNX_MESSAGE_IPC */
 } DltReceiver;
 
 typedef struct
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -74,6 +74,23 @@
 #include "dlt_daemon_offline_logstorage.h"
 #include "dlt_gateway.h"
 
+#ifdef __QNX__
+#   include <errno.h>
+#   include <sys/neutrino.h>
+#   include <sys/types.h>
+#   include <sys/iofunc.h>
+#   include <sys/dispatch.h>
+#   ifdef DLT_DAEMON_USE_QNX_MESSAGE_IPC
+        static pthread_t t1;
+        bool work_in_progress = true;
+        struct msg_thread_arg {
+            DltDaemon* daemon;
+            DltDaemonLocal* daemon_local;
+        };
+        void* msg_thread(void* arg);
+#   endif /* DLT_DAEMON_USE_QNX_MESSAGE_IPC */
+#endif /* __QNX__ */
+
 #ifdef UDP_CONNECTION_SUPPORT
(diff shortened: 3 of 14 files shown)


CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ option(WITH_DLT_TESTS         "Set to ON to build src/test binaries"
 option(WITH_DLT_UNIT_TESTS    "Set to ON to build gtest framework and tests/binaries"                            OFF)
 option(WITH_DLT_QNX_SYSTEM    "Set to ON to build QNX system binary dlt-qnx-system"                              OFF)
 
-set(DLT_IPC "FIFO" CACHE STRING "UNIX_SOCKET,FIFO")
+set(DLT_IPC "FIFO" CACHE STRING "UNIX_SOCKET,FIFO,QNX_MESSAGE")
```
ssugiura, line 70: If the user set to QNX_MESSAGE, we would verify first that it is compiled under QNX, otherwise raise an error and return already. You could do something similar to line 161. Then you don't need to consider QNX and non-QNX inside headers and codes.

include/dlt/dlt_common.h
```diff
@@ -111,6 +111,16 @@
 #      define DLT_PACKED
 #   endif
 
+#   ifdef DLT_DAEMON_USE_QNX_MESSAGE_IPC
+#      ifdef __QNX__
```
ssugiura, line 115: As commented above, we don't require this ifdef, no error message at line 120 either. The same for other places.

src/daemon/dlt-daemon.c
```diff
@@ -74,6 +74,23 @@
 #include "dlt_daemon_offline_logstorage.h"
 #include "dlt_gateway.h"
 
+#ifdef __QNX__
+#   include <errno.h>
+#   include <sys/neutrino.h>
+#   include <sys/types.h>
+#   include <sys/iofunc.h>
+#   include <sys/dispatch.h>
```
ssugiura, line 82: We need this only for DLT_DAEMON_USE_QNX_MESSAGE_IPC, I guess

src/daemon/dlt_daemon_common.c
```diff
@@ -85,6 +85,19 @@
...
+        iofunc_attr_t    attr;
+        DltReceiver      *receiver;
+    };
+    #define IOFUNC_ATTR_T       struct extattr
+#endif /* DLT_DAEMON_USE_QNX_MESSAGE_IPC */
+#ifdef __QNX__
+#   include <sys/neutrino.h>
+#   include <sys/iofunc.h>
+#   include <sys/dispatch.h>
+#endif /* __QNX__ */
```
ssugiura, line 99: Do we need this even if QNX_MESSAGE_IPC is not set?

src/lib/dlt_user.c
```diff
@@ -259,7 +262,44 @@ static DltReturnValue dlt_socket_set_nonblock_and_linger(int sockfd)
...
+            dlt_user_initialised = false;
+            rc = DLT_RETURN_ERROR;
+        }
+    }
+
+    if (dlt_user.receiver.attach == NULL && dlt_attach_channel(&attach) == DLT_RETURN_OK) {
+        dlt_user.receiver.attach = attach;
+    }
+
+DLT_SEM_FREE();
```
ssugiura, line 298: Correct indent

src/lib/dlt_user.c
```diff
@@ -259,7 +262,44 @@ static DltReturnValue dlt_socket_set_nonblock_and_linger(int sockfd)
...
 
-#ifdef DLT_LIB_USE_UNIX_SOCKET_IPC
+#ifdef DLT_LIB_USE_QNX_MESSAGE_IPC
+static DltReturnValue dlt_initialize_resmgr_connection(void)
+{
+    DltReturnValue rc = DLT_RETURN_OK;
+    int fd = -1;
+    name_attach_t *attach;
+
+DLT_SEM_LOCK();
```
ssugiura, line 272: Correct indent

ssugiura (changes requested):
Overall implementation looks fine with me.
I also would like to request you to document overall design and sequence of QNX messaging.
Could you prepare a new document with e.g. `doc/dlt_ipc_qnx_message.md` and describe this?
Afterwards I'd like to add documents for other IPCs (FIFO and unix socket).
This would definitely help new comers to decide which IPC to be used.
Thanks in advance!

minminlittleshrimp:
Hello @svstanko 

Do you still have any interest in this topic?
If yes could you please let me know if you could resolve the conflict so we could continue reviewing for you?

Thank you

michael-methner:
Closing pull request as there seems to be no interest of the author anymore.
Please reopen if you want to rework it.

alexmohr:
@marcin-sochacki As you're employed by BlackBerry it would be awesome if you pickup the slack here. Getting message passing to work fully reduces the load in dlt-daemon tremendously. 
We did some work internally to improve on this MR and it showed reductions in CPU usage of up to 30%. Unfortunately we never got it fully running. 

Imho this would be huge win in the context of "QNX everywhere" :wink:

marcin-sochacki:
@alexmohr @minminlittleshrimp @michael-methner 
This is indeed a great performance improvement!
I would like to understand if dlt-daemon is going to be used in the future automotive projects (and if that optimization would help the customers) or is t just maintained for existing projects (bug fixes, etc.).
This "Pull Request" was created in 2021 and the code base changed a lot since then, so I expect that rebasing may not be a trivial task.
Anyway I could look at it, but the precondition is another pull request with build system changes for QNX:
https://github.com/COVESA/dlt-daemon/pull/609

alexmohr:
Obviously I can't speak too much about in which direction development is going for us, but for the foreseeable future dlt-daemon is likely to be a central component in the logging architecture. 
Even for current generations getting this improvement, would be a huge benefit for us and probably everyone else who is using dlt-daemon on QNX.

michael-methner:
Hello @marcin-sochacki ,
i cannot disclose any details on our current projects but I would second @alexmohr statements. Having this performance improvements on QNX would be great as DLT on QNX is often used as central logging component in multi-os setups.

svstanko:
Hello @michael-methner,
Sorry for delay with the answer.
That solution was abandoned after deeper investigation. It passed all our tests but failed with more complicated scenarios of other components.
Tests failed due to unexpected stopping of dlt-deamon in some test cases with stopping several components that uses dlt. (Stopping only component didn't affect dlt in such way).
The root cause of the issue:
When the daemon receives a request to set (change) the log level, it tries to send this request to all clients, and when trying to send such a request to clients that have already disconnected, an error occurs, according to which the daemon exits its event processing loop.
The solution hasn't been found.

marcin-sochacki:
@alexmohr 
QNX would be happy to help.
So far I re-based the commit onto recent software baseline, applied the changes from the code review and fixed a dependency issue, which didn't allow to build unit tests for QNX:
https://github.com/marcin-sochacki/dlt-daemon/tree/resource_manager
I run unit tests under QNX in QEMU and some of them fail.
Since this is my background task, it may take some time, but I'm going to investigate the issues and try to fix them. Until now, you may try it out from my fork.

alexmohr:
@marcin-sochacki That is great news, thank you! 

Can you post which tests are failing? We are running unit tests on qnx internally as well but did some patches to the tests to fix some issues. So maybe we fixed these issues already and can upstream the required changes.

We never bothered to upstream so far these because the changes can't be validated in the github pipeline anyways.

marcin-sochacki:
Here are the test reports: 
[test_reports.zip](https://github.com/COVESA/dlt-daemon/files/15309115/test_reports.zip)

alexmohr:
We actually had to implement quite some changes to get these to run. 
https://github.com/mercedes-benz/dlt-daemon/commit/f5de8907ef8080ec601185386fb031efdcc72a70

marcin-sochacki:
> Our issue with this was that clients running with a higher priotity than dlt-daemon get degraded to the dlt-daemon priotity while sending messages. Through the priotity inversion cricitical processes don't get CPU time anymore (or too late) resulting in a watchdog kill.

QNX uses priority inheritance to actually prevent problems of high priority clients blocked on servers servicing low priority clients. QNX boost the server's priority to be the highest of all blocked clients:
https://www.qnx.com/developers/docs/7.1/#com.qnx.doc.neutrino.getting_started/topic/s1_msg_prio_interitance.html
However there are some exceptions, for example if threads use semaphores (instead of mutexes) for synchronization.
I will take a look at the above changes. As mentioned before, I only re-based the commit onto recent software baseline and fixed obvious compilation issues, but didn't look at the content so far.

marcin-sochacki:
I compared:
https://github.com/marcin-sochacki/dlt-daemon/tree/resource_manager
with
https://github.com/mercedes-benz/dlt-daemon/tree/message-passing
and it looks that your branch contains also other changes, making it more difficult to have a common codebase.

My quick findings:
1) dlt_receiver_receive()
function may not always reply to client's MsgSend(), keeping them blocked.
There are several if else statements there and MsgReply() is called only for one condition.
In other cases, the client should also get an answer - MsgError() should be called to unblock the client.
2) The Resource Manager shall implement close_dup handler to clean-up client-oriented resources,
otherwise the Resource Manager does not know if a client died or simply closed the connection.
3) And as already mentioned: Semaphores don't affect a thread's effective priority; if you need priority inheritance, use a mutex.

I hope that helps.

PR #351, opened 2021-11-13
Support for Cygwin toolchain.

DoctorNoobingstoneIPresume:
Also tested with the latest version running on Windows XP (gcc-5.4.0)
(http://www.crouchingtigerhiddenfruitbat.org/cygwin/timemachine.html).

### DIFF
--- a/src/daemon/CMakeLists.txt
+++ b/src/daemon/CMakeLists.txt
@@ -45,7 +45,7 @@ if(WITH_DLT_SHM_ENABLE)
         ${PROJECT_SOURCE_DIR}/src/shared/dlt_shm.c)
 endif()
 
-if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+if("${CMAKE_SYSTEM_NAME}" MATCHES "Linux|CYGWIN")
     set(RT_LIBRARY rt)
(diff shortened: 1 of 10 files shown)


DoctorNoobingstoneIPresume:
It seems that the `open` function is declared in `<fcntl.h>`, not in `<unistd.h>` (https://www.man7.org/linux/man-pages/man2/open.2.html).

DoctorNoobingstoneIPresume:
It seems that for the Cygwin toolchain, just like the Linux toolchain, we should not pass `-lsocket` to the linker.

DoctorNoobingstoneIPresume:
Regarding CMake scripts:

I wonder why quotes are needed around `${CMAKE_SYSTEM_NAME}` for the `MATCHES` operator, but not for the `STREQUAL` operator. (I could see that already we have been using quotes with `MATCHES`. I could also see that, without quotes, `MATCHES` does not seem to work as expected.)

Anyway, maybe it is useful (good practice, good model) to use quotes (almost) everywhere around variable names. For example, should/might we replace `${CMAKE_SYSTEM_NAME}` with `"${CMAKE_SYSTEM_NAME}"` everywhere ?

PR #362, opened 2022-01-19
dlt-user: Fix crashes in dlt_free during dlt_init

alexmohr:
If dlt_free is called while dlt_init is running, the dlt-user library
crashes because the init and free where not thread safe.
This commit introduces a new field dlt_user_initialising
which will prevent entering dlt_free while dlt_init is running.
Correctness is ensured by a unit test

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -90,9 +90,17 @@
 #   define DLT_LOG_FATAL_RESET_TRAP(LOGLEVEL)
 #endif /* DLT_FATAL_LOG_RESET_ENABLE */
 
+enum InitState {
+    INIT_UNITIALIZED,
+    INIT_IN_PROGRESS,
+    INIT_DONE
+};
+
 static DltUser dlt_user;
-static atomic_bool dlt_user_initialised = false;
-static int dlt_user_freeing = 0;
+static _Atomic enum InitState dlt_user_init_state = INIT_UNITIALIZED;
+#define DLT_USER_INITALIZED (dlt_user_init_state == INIT_DONE)
+
+static _Atomic int dlt_user_freeing = 0;
 static bool dlt_user_file_reach_max = false;
 
 #ifdef DLT_LIB_USE_FIFO_IPC
@@ -308,7 +316,7 @@ static DltReturnValue dlt_initialize_socket_connection(void)
                               sockfd,
                               DLT_RECEIVE_SOCKET,
                               DLT_USER_RCVBUF_MAX_SIZE) == DLT_RETURN_ERROR) {
-            dlt_user_initialised = false;
+            dlt_user_init_state = INIT_UNITIALIZED;
             close(sockfd);
             DLT_SEM_FREE();
             return DLT_RETURN_ERROR;
@@ -363,7 +371,7 @@ static DltReturnValue dlt_initialize_vsock_connection()
                               sockfd,
                               DLT_RECEIVE_SOCKET,
                               DLT_USER_RCVBUF_MAX_SIZE) == DLT_RETURN_ERROR) {
-            dlt_user_initialised = false;
+            dlt_user_init_state = INIT_UNITIALIZED;
             close(sockfd);
             DLT_SEM_FREE();
             return DLT_RETURN_ERROR;
@@ -445,25 +453,6 @@ static DltReturnValue dlt_initialize_fifo_connection(void)
 
 DltReturnValue dlt_init(void)
 {
-    /* Compare 'dlt_user_initialised' to false. If equal, 'dlt_user_initialised' will be set to true.
-    Calls retruns true, if 'dlt_user_initialised' was false.
-    That way it's no problem, if two threads enter this function, because only the very first one will
-    pass fully. The other one will immediately return, because when it executes the atomic function
-    'dlt_user_initialised' will be for sure already set to true.
-    */
-    bool expected = false;
-    if (!(atomic_compare_exchange_strong(&dlt_user_initialised, &expected, true)))
-        return DLT_RETURN_OK;
-
-    /* check environment variables */
-    dlt_check_envvar();
-
-    /* Check logging mode and internal log file is opened or not*/
-    if(logging_mode == DLT_LOG_TO_FILE && logging_handle == NULL)
-    {
-        dlt_log_init(logging_mode);
-    }
-
     /* process is exiting. Do not allocate new resources. */
     if (dlt_user_freeing != 0) {
         dlt_vlog(LOG_INFO, "%s logging disabled, process is exiting", __func__);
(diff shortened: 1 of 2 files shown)


src/lib/dlt_user.c
```diff
@@ -90,8 +90,16 @@
...
 
+enum InitState {
+    INIT_UNITIALIZED,
+    INIT_IN_PROGRESS,
+    INIT_DONE
+};
+
 static DltUser dlt_user;
-static atomic_bool dlt_user_initialised = false;
+static enum InitState dlt_user_init_state = INIT_UNITIALIZED;
```
mbehr1, line 100: shouldn't this be "static volatile enum ..."  as you're later using atomic_compare_exchange_strong with it?

alexmohr: Yes good point, added it :+1:

src/lib/dlt_user.c
```diff
@@ -445,35 +453,37 @@ static DltReturnValue dlt_initialize_fifo_connection(void)
...
+     * That way it's no problem, if two threads enter this function, because only the very first one will
+     * pass fully. The other one will immediately return, because when it executes the atomic function
+     * dlt_user_init_state won't be INIT_UNITIALIZED anymore.
+     * This is not handled via a simple boolean to prevent issues with shutting down while the init is still running.
+     * Furthermore, this makes sure we enter some function only when dlt_init is fully done.
+     * */
+    enum InitState expectedInitState = INIT_UNITIALIZED;
+    if (!(atomic_compare_exchange_strong(&dlt_user_init_state, &expectedInitState, INIT_IN_PROGRESS))) {
         return DLT_RETURN_OK;
+    }
```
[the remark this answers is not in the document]
alexmohr: Hi @michael-methner the described scenario could leave the library in an initialized state.
Adding a mutex here sounds like a good idea. I'll add it later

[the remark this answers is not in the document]
alexmohr: done :)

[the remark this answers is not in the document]
alexmohr: @michael-methner It's done in case you did not notice it yet :)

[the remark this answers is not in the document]
michael-methner: Hello @alexmohr ,
sorry for the delay and thank you very much for the rework. I was quite busy the last weeks. I need some time to carefully review this as it is related to race conditions ;). I hope I can manage before end of this week.

One first question: Would it make sense to use the same mutex in dlt_free() as well to avoid race conditions between dlt_init() and dlt_free().

src/lib/dlt_user.c
```diff
@@ -90,8 +90,17 @@
 #   define DLT_LOG_FATAL_RESET_TRAP(LOGLEVEL)
 #endif /* DLT_FATAL_LOG_RESET_ENABLE */
 
+enum InitState {
```
lvklevankhanh, line 93: Nice to have if the name of enum should be DltInitState

src/lib/dlt_user.c
```diff
@@ -445,36 +454,42 @@ static DltReturnValue dlt_initialize_fifo_connection(void)
...
+     * That way it's no problem, if two threads enter this function, because only the very first one will
+     * pass fully. The other one will immediately return, because when it executes the atomic function
+     * dlt_user_init_state won't be INIT_UNITIALIZED anymore.
+     * This is not handled via a simple boolean to prevent issues with shutting down while the init is still running.
+     * Furthermore, this makes sure we enter some function only when dlt_init is fully done.
+     * */
+    enum InitState expectedInitState = INIT_UNITIALIZED;
+    if (!(atomic_compare_exchange_strong(&dlt_user_init_state, &expectedInitState, INIT_IN_PROGRESS))) {
+        /* skip init if we're initialized, otherwise try init (again) */
+        if (dlt_user_init_state == INIT_DONE)
```
lvklevankhanh, line 478: Nice to have: the condition should be "if (DLT_USER_INITALIZED)"

src/lib/dlt_user.c
```diff
@@ -1735,7 +1762,7 @@ int dlt_set_resend_timeout_atexit(uint32_t timeout_in_milliseconds)
     if (g_dlt_is_child)
         return DLT_RETURN_ERROR;
 
-    if (dlt_user_initialised == 0)
+    if (DLT_USER_INITALIZED == 0)
```
lvklevankhanh, line 1765: Nice to have: the condition should be "if (!DLT_USER_INITALIZED)"

src/lib/dlt_user.c
```diff
@@ -530,14 +544,25 @@ DltReturnValue dlt_init(void)
...
     if (dlt_start_threads() < 0) {
-        dlt_user_initialised = false;
-        return DLT_RETURN_ERROR;
+        goto return_error;
     }
 
     /* prepare for fork() call */
     pthread_atfork(NULL, NULL, &dlt_fork_child_fork_handler);
 
+    /* set success before jump label because in case we used the jump INIT_DON
```
lvklevankhanh, line 553: Nice to have: "INIT_DONE" instead "INIT_DON"

alexmohr:
@lvklevankhanh thanks for your comments I'm going to fix as soon as I'm back home. It will take a couple of weeks though. Just saying so you don't close the PR due to inactivity :)

src/lib/dlt_user.c
michael-methner: This needs to be:
static _Atomic enum InitState dlt_user_init_state = INIT_UNITIALIZED;
and also:
static _Atomic int dlt_user_freeing = 0;

otherwise it does not compile with clang anymore.

src/lib/dlt_user.c
```diff
@@ -90,8 +90,17 @@
 #   define DLT_LOG_FATAL_RESET_TRAP(LOGLEVEL)
 #endif /* DLT_FATAL_LOG_RESET_ENABLE */
 
+enum InitState {
+    INIT_UNITIALIZED,
```
michael-methner, line 94: typo: INIT_UNINTIALIZED

src/lib/dlt_user.c
```diff
@@ -90,8 +90,17 @@
...
+    INIT_IN_PROGRESS,
+    INIT_DONE
+};
+
 static DltUser dlt_user;
-static atomic_bool dlt_user_initialised = false;
+static volatile enum InitState dlt_user_init_state = INIT_UNITIALIZED;
+static pthread_mutex_t dlt_user_init_mtx = PTHREAD_MUTEX_INITIALIZER;
+#define DLT_USER_INITALIZED (dlt_user_init_state == INIT_DONE)
```
michael-methner, line 102: typo: DLT_USER_INITIALIZED

src/lib/dlt_user.c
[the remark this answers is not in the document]
alexmohr: Thanks for the feedback. Both variables are changed to `_Atomic` and it is building locally with clang 10 and gcc 10.3

PR #369, opened 2022-03-02
logfile: exhance internal dlt logging by introducing size limits

danielweber2018:
Enhance dlt logging such that multiple files are used as it is done
for the offline traces. Add limit-specific config values for logging.
For this purpose the pattern of index-based file names is used only.
This approach of logging to multiple files and rotating in order
to keep the limits ensures that dlt logs take care of available
space on the underlying file system and do not grow infinitely.

<sup>Daniel Weber, <daniel.w.weber@daimler.com>, Daimler TSS GmbH, [imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sup>


danielweber2018:
Hi @minminlittleshrimp 
From my side there is definitely still an interest in implementing this feature.

Maybe for a short explanation of my changes:
For my PR I used the offline trace pattern (size limit with multiple files, the files rotates via an file-counter, all configurable via config file). It was important to me to avoid duplicate code, so I refactored out the relevant code blocks from the offline trace. Unfortunately, that bloated the PR a bit.

I haven't maintained the PR (e.g. the new conflicts), but if there is interest in this PR I'll be happy to take care of it. If you have any questions, please contact me.

Thanks and regards Daniel

minminlittleshrimp:
Hello @danielweber2018 
Thank you for your information.
Then it would be very great if you could maintain your PR.
Thank you very much and looking forward to cooperating with you!
Minh

include/dlt/dlt_common.h
```diff
@@ -88,20 +89,13 @@
 #      include <time.h>
 #   endif
 
-#   if defined(__GNUC__)
```
michael-methner, line 91: Changes to dlt_common.h (and all other header files in this folder) should be treated as API changes. Therefore this patch would require another major version release of DLT.

is there a way to keep the header files untouched for this feature?

src/shared/dlt_log.c
```diff
@@ -0,0 +1,352 @@
...
+        .filename={0},
+        .fileSize=0,
+        .maxSize=0,
+        .filenameTimestampBased=false,
+        .filenameBase={0},
+        .filenameExt={0},
+        .ohandle=-1};
+
+
+void dlt_log_set_filename(const char *filename)
```
michael-methner, line 57: Why is it really necessary  to move all this code from dlt_common.c to dlt_log.c? These cleanup activities make it hard to review the change it the PRs get really big. Two seperate PRs here: one for the clean up / code moving. And a second one which implements the feature.

include/dlt/dlt_common.h
danielweber2018: Hi @michael-methner ,
I don't think we can do this feature completely untouched. But we made sure that we are 100% compatible with the old header (otherwise we would have had various build problems in projects using our DLT header).
As requested by the other comment, I would remove the movement from dlt_common.h to dlt_log.h. I hope then the compatibility becomes more visible.

src/shared/dlt_log.c
danielweber2018: Of course, the split between dlt_common and dlt_log is not mandatory for the feature. I would remove the movement from dlt_common to dlt_common.
If my feature PR is merged, I will create another PR that contains the split.

danielweber2018:
I rebased my PR to the current master and resolved all conflicts. Additionally i simplified the PR via the removing of the split between dlt_common and dlt_log (like requests from michael methner).

src/shared/dlt_common.c
```diff
@@ -1815,35 +1828,115 @@ void dlt_print_with_attributes(bool state)
...
-        logging_handle = fopen(logging_filename, "a");
+    if (logging_mode != DLT_LOG_TO_FILE) {
+        return DLT_RETURN_OK;
+    }
 
-        if (logging_handle == NULL) {
-            dlt_user_printf("Internal log file %s cannot be opened!\n", logging_filename);
-            return DLT_RETURN_ERROR;
+    if (enable_multiple_logfiles) {
+        dlt_user_printf("configure dlt logging using file limits\n");
```
michael-methner, line 1850: Sorry for let you waiting, but the amount of changes really take time to review. I have tested your changes and they work fine on my systems. Only thing I noticed is that above log message is printed twice. It is called from these two locations:
First call:
#0  dlt_log_init_multiple_logfiles_support (mode=DLT_LOG_TO_FILE, enable_multiple_logfiles=true, logging_file_size=200, logging_files_max_size=1000)
    at dlt-daemon/src/shared/dlt_common.c:1837
#1  0x000055555555d624 in main (argc=1, argv=0x7fffffffdb78) at /home/mmethner/src/dlt_size_limits/dlt-daemon/src/daemon/dlt-daemon.c:960


Second call:
#0  dlt_log_init_multiple_logfiles_support (mode=DLT_LOG_TO_FILE, enable_multiple_logfiles=true, logging_file_size=200, logging_files_max_size=1000)
    at dlt-daemon/src/shared/dlt_common.c:1837
#1  0x00005555555617c3 in dlt_daemon_local_init_p1 (daemon=<optimized out>, daemon_local=0x7fffffff7370, verbose=<optimized out>)
    at dlt-daemon/src/daemon/dlt-daemon.c:1198
#2  0x000055555555d792 in main (argc=<optimized out>, argv=<optimized out>)
    at dlt-daemon/src/daemon/dlt-daemon.c:1007
(gdb)

danielweber2018: Hi Michael,
thanks for the response :)

I use the already included code paths. As far as I understood the existing code this codeblock init the early log (e.g. for some startup logs) https://github.com/COVESA/dlt-daemon/blob/master/src/daemon/dlt-daemon.c#L944 and then reinit after forking https://github.com/COVESA/dlt-daemon/blob/master/src/daemon/dlt-daemon.c#L1171 . The log initializations are therefore executed twice as before (and create so the duplicated log line).

Is this pattern wrong? Is the logline really a problem? If yes, do you have a suggestion how I should restructure it?

Thanks in advanced

michael-methner: Hello @danielweber2018 ,
i did not see any concrete issues (like memory leaks) or so due to the double call to this function.

As you correctly said, the behavior was there even before your patch, so I will continue with merging this patch and I have created an issue to track the double initialization (#452 ).

Thank you very much for your contribution and your long lasting support to get this patch merged :+1:

tests/gtest_dlt_daemon_multiple_files_logging.cpp
```diff
@@ -0,0 +1,277 @@
...
+ * unnecessary two files. The reinit have to append to the first file.
+ */
+TEST(t_dlt_logging_multiple_files_append_reinit, normal)
+{
+    const char* path = "/tmp";
+    const char* file_name = "dlt.log";
+    const int file_size = 256;
+    const int max_file_size = 512;
+
+    const char* log1 = "ONE\n";
```
minminlittleshrimp, line 102: Hello @danielweber2018 
Build logs show me:
```
Scanning dependencies of target gtest_dlt_daemon_multiple_files_logging
[ 86%] Building CXX object tests/CMakeFiles/gtest_dlt_daemon_multiple_files_logging.dir/gtest_dlt_daemon_multiple_files_logging.cpp.o
/home/lum3hc/work/github/PR442/dlt-daemon/tests/gtest_dlt_daemon_multiple_files_logging.cpp: In member function ‘virtual void t_dlt_logging_multiple_files_append_reinit_normal_Test::TestBody()’:
/home/lum3hc/work/github/PR442/dlt-daemon/tests/gtest_dlt_daemon_multiple_files_logging.cpp:106:28: warning: format not a string literal and no format arguments [-Wformat-security]
     dlt_vlog(LOG_INFO, log1);
                            ^
/home/lum3hc/work/github/PR442/dlt-daemon/tests/gtest_dlt_daemon_multiple_files_logging.cpp:110:28: warning: format not a string literal and no format arguments [-Wformat-security]
     dlt_vlog(LOG_INFO, log2);
                            ^
[ 87%] Linking CXX executable gtest_dlt_daemon_multiple_files_logging
```
Could you please consider to use string literal format?
Maybe char array?

src/shared/dlt_multiple_files.c
```diff
@@ -0,0 +1,499 @@
...
+                         files_buffer->directory, newest);
+
+    if ((ret < 0) || (ret >= NAME_MAX)) {
+        fprintf(stderr, "filename cannot be concatenated\n");
+        return DLT_RETURN_ERROR;
+    }
+
+    /* open DLT output file */
+    errno = 0;
+    files_buffer->ohandle = open(file_path, O_WRONLY | O_APPEND, S_IRUSR | S_IWUSR |
```
minminlittleshrimp, line 402: Hello @danielweber2018 
I would like to remove the superfluous mode bits in PR: https://github.com/COVESA/dlt-daemon/pull/462
Thank you

PR #373, opened 2022-04-18
daemon & lib: add path to custom unix socket file

vsloginov:
- Add param to the DLT daemon to set custom socket file
- Add env var to the DLT library to set custom socket file


src/daemon/dlt-daemon.c
```diff
@@ -259,6 +271,14 @@ int option_handling(DltDaemonLocal *daemon_local, int argc, char *argv[])
         }
 #endif
 
+#ifdef DLT_DAEMON_USE_UNIX_SOCKET_IPC
+        case 'u':
+        {
+            strncpy(daemon_local->flags.appSockPath, optarg, NAME_MAX);
```
michael-methner, line 277: Could you please use DLT_IPC_PATH_MAX here instead of NAME_MAX.

src/daemon/dlt-daemon.c
```diff
@@ -377,14 +397,7 @@ int option_file_parser(DltDaemonLocal *daemon_local)
     strncpy(daemon_local->flags.ctrlSockPath,
             DLT_DAEMON_DEFAULT_CTRL_SOCK_PATH,
             sizeof(daemon_local->flags.ctrlSockPath));
-#ifdef DLT_DAEMON_USE_UNIX_SOCKET_IPC
-    snprintf(daemon_local->flags.appSockPath, DLT_IPC_PATH_MAX, "%s/dlt", DLT_USER_IPC_PATH);
-
-    if (strlen(DLT_USER_IPC_PATH) > DLT_IPC_PATH_MAX)
```
michael-methner, line 383: Could you please add this check again for both cases (using DLT_USER_IPC_PATH and optarg).

src/daemon/dlt-daemon.c
vsloginov: ok

src/daemon/dlt-daemon.c
vsloginov: Sorry, this commit was made quite long time ago, so I could be a bit out of scope. 
As I see we dont need here additional check, because we are using above s**n**printf() and str**n**cpy() which are limited by DLT_IPC_PATH_MAX. 
If this statement is correct, could you please elaborate why do you think those checks are needed here?

michael-methner: Hello @vsloginov ,
sorry for the delay. I some how missed your message.
This check is to print a warning to stderr in case snprintf or strncpy is trimming the path due to size limites.

If I think about the if statement above, I think it is even not correct and has to be:
if(strlen(DLT_USER_IPC_PATH) +4 > DLT_IPC_PATH_MAX)
as we add the four characters "/dlt" to the string.

Does this make sense?

michael-methner:
Hello @vsloginov ,
there are two comments which require rework. Are you willing to do the rework? Otherwise we will close this PR without merging.
Thank you for your understanding.

michael-methner:
No rework done. Please feel free to reopen once rework was done.

PR #381, opened 2022-05-24
dlt-gateway: Fix crash on invalid ip

alexmohr:
* dlt-gateway crashed when an invalid
  ip address was configured as a null pointer
  was used for strdup

* Add timeout for dlt-client-connect.
  Timeout is hard coded to 1 seconds to prevent blocking
  the dlt-daemon overly long if a timeout is happening
  while creating a gateway connection.

### DIFF
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -1045,7 +1045,7 @@ int main(int argc, char *argv[])
     /* initiate gateway */
     if (daemon_local.flags.gatewayMode == 1) {
         if (dlt_gateway_init(&daemon_local, daemon_local.flags.vflag) == -1) {
-            dlt_log(LOG_CRIT, "Fail to create gateway\n");
+            dlt_log(LOG_CRIT, "Failed to create gateway\n");
             return -1;
         }
 
--- a/src/gateway/dlt_gateway.c
+++ b/src/gateway/dlt_gateway.c
@@ -791,6 +791,10 @@ int dlt_gateway_configure(DltGateway *gateway, char *config_file, int verbose)
                              configuration_entries[j].key, value);
             }
 
+            if (!tmp.ip_address) {
+                invalid = 1;
+            }
+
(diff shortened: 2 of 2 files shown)


thanhbnq:
Hello @alexmohr ,
Thank you for your PR.

Regarding the issue with dlt_client_connet, we also implemented the fix and I would like to get your feedback on https://github.com/COVESA/dlt-daemon/pull/409

About the first issue "dlt-gateway crashed when an invalid ip address was configured as a null pointer was used for strdup",
I am not really sure your fix will help to solve it. I mean in the check dlt_gateway_check_ip(), we already prevent null pointer already.

Best regards,
Thanh

alexmohr:
> Regarding the issue with dlt_client_connet, we also implemented the fix and I would like to get your feedback on #409
Looks like a valid alternative to what I'm suggesting here, but personally I find setting a timeout to the socket easier to maintain and read compared to setting / unsetting nonblock mode.

> About the first issue "dlt-gateway crashed when an invalid ip address was configured as a null pointer was used for strdup", I am not really sure your fix will help to solve it. I mean in the check dlt_gateway_check_ip(), we already prevent null pointer already.

Yes and no the important change here is 
```
 if (!tmp.ip_address) {
                invalid = 1;
            }
```
without setting the ip address to invalid the daemon will crash.

thanhbnq:
> Looks like a valid alternative to what I'm suggesting here, but personally I find setting a timeout to the socket easier to maintain and read compared to setting / unsetting nonblock mode.

In some cases, the connection is in progress of establishment and setting timeout could be not getting the exact result.
So if you are fine with our solution, then please kindly update this PR for dlt-gateway fix. I believe that we can apply your fix straightforwardly.

Thank you.
Thanh

alexmohr:
This PR now only contains the fix for invalid gateway addresses and a spelling fix. as #409 supersedes the timeout implementation 

Thanks
Alex

PR #385, opened 2022-06-15
dlt_user_shared: Add timeout to writev

alexmohr:
This timeout is necessary to prevent
blocking writev indefinitely.
Without the timeout dlt-daemon, may block
indefinitely when an app id is re-used
very frequently.
In that case dlt-daemon won't accept anymore
new connections and further communication
in any way is not possible anymore.

### DIFF
--- a/src/shared/dlt_user_shared.c
+++ b/src/shared/dlt_user_shared.c
@@ -71,6 +71,7 @@
 #include <errno.h>
 
 #include <sys/uio.h> /* writev() */
+#include <sys/time.h> /* timeval */
 
 #include "dlt_user_shared.h"
 #include "dlt_user_shared_cfg.h"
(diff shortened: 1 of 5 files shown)


src/shared/dlt_user_shared.c
```diff
@@ -175,4 +198,26 @@ DltReturnValue dlt_user_log_out3(int handle, void *ptr1, size_t len1, void *ptr2
...
+{
+    if (handle < 0)
+        /* Invalid handle */
+        return DLT_RETURN_ERROR;
+
+    fd_set fds;
+    FD_ZERO(&fds);
+    FD_SET(handle, &fds);
+
+    struct timeval tv = { 0, 100000000 };
```
michael-methner, line 213: 100000000 usec = 100 sec.
Isn't this a bit to much?
Please add a configurable value for all these values. Is 10 seconds sufficient as default value? This would be more readable:
struct timeval tv = { 10, 0 };

alexmohr: Good catch, I checked our internal fork and I somehow screwed up while applying this patch here. We're actually using 1 second timeout and not 100. 
I adjusted the value to reflect this. 
Also made it configurable via CMake

michael-methner:
Hello @alexmohr ,
i have tested your changes and working fine with me as well.

But I am still a bit hesitant to merge it as it might cause problems when a system is under high cpu load and timeout of 1 sec is triggered by accident.

Can you please describe how to reproduce the issue?

You stated above "indefinitely when an app id is re-used very frequently." Does this mean the app-id is registered from the same process several times? When it is registered from the different process, it should not matter as each process uses its own fifo.

alexmohr:
> Hello @alexmohr , i have tested your changes and working fine with me as well.
> 
> But I am still a bit hesitant to merge it as it might cause problems when a system is under high cpu load and timeout of 1 sec is triggered by accident.
> 
> Can you please describe how to reproduce the issue?
> 
> You stated above "indefinitely when an app id is re-used very frequently." Does this mean the app-id is registered from the same process several times? When it is registered from the different process, it should not matter as each process uses its own fifo.

michael-methner (approved):

michael-methner:
I gave another try to reproduce the behavior but failed. Nevertheless, I think the patch will be an improvement for these high-load scenarios -> merging.

@alexmohr : Thank you for the contribution.

PR #396, opened 2022-08-01
Logging to file: Add config option to overwrite log file

denysbalym:
If dlt-daemon is configured to log to file, the log file will grow unlimited because it is always appended.

### DIFF
--- a/include/dlt/dlt_common.h
+++ b/include/dlt/dlt_common.h
@@ -1194,8 +1194,17 @@ void dlt_print_with_attributes(bool state);
 /**
  * Initialize (external) logging facility
  * @param mode positive, 0 = log to stdout, 1 = log to syslog, 2 = log to file, 3 = log to stderr
+ * @param overwrite_flag 1 = overwrite LoggingFilename, 0 = append LoggingFilename
  */
-void dlt_log_init(int mode);
+void dlt_log_init_p1(int mode, int overwrite_flag);
+
+/**
+ * Reinitialize (external) logging facility
+ * Should be called after dlt_log_init_p1() to reinitialize after fork
+ * @param mode positive, 0 = log to stdout, 1 = log to syslog, 2 = log to file, 3 = log to stderr
+ */
+void dlt_log_init_p2(int mode);
+
 /**
  * Print with variable arguments to specified file descriptor by DLT_LOG_MODE environment variable (like fprintf)
  * @param format format string for message
--- a/src/shared/dlt_common.c
+++ b/src/shared/dlt_common.c
@@ -1814,7 +1814,7 @@ void dlt_print_with_attributes(bool state)
     print_with_attributes = state;
 }
 
-void dlt_log_init(int mode)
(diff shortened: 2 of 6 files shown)


include/dlt/dlt_common.h
```diff
@@ -1194,8 +1194,17 @@ void dlt_print_with_attributes(bool state);
 /**
  * Initialize (external) logging facility
  * @param mode positive, 0 = log to stdout, 1 = log to syslog, 2 = log to file, 3 = log to stderr
+ * @param overwrite_flag 1 = overwrite LoggingFilename, 0 = append LoggingFilename
  */
-void dlt_log_init(int mode);
+void dlt_log_init_p1(int mode, int overwrite_flag);
```
minminlittleshrimp, line 1199: Hello dbalym,
Could you tell me the reason why we need to separate the primitive API dlt_log_init into 2 others?
I am not so clear about this change.
Thank you

src/shared/dlt_common.c
```diff
@@ -1825,7 +1825,7 @@ void dlt_log_init(int mode)
 
     if (logging_mode == DLT_LOG_TO_FILE) {
         /* internal logging to file */
-        logging_handle = fopen(logging_filename, "a");
+        logging_handle = fopen(logging_filename, file_mode);
```
minminlittleshrimp, line 1828: I think we could create an internal function, something like const char *dlt_set_log_to_file_mode(int overwrite_flag) instead of trying to add 1 more argument.
How do you think?

src/shared/dlt_common.c
```diff
@@ -3992,7 +4006,7 @@ void dlt_check_envvar()
         int mode = 0;
 
         if (sscanf(env_log_mode, "%d", &mode))
-            dlt_log_init(mode);
+            dlt_log_init_p2(mode);
```
minminlittleshrimp, line 4009: I try to build with your change, but it throws error:
[ 40%] Linking C executable dlt-convert
../lib/libdlt.so.2.18.8: undefined reference to `dlt_log_init'
collect2: error: ld returned 1 exit status
src/console/CMakeFiles/dlt-convert.dir/build.make:96: recipe for target 'src/console/dlt-convert' failed
make[2]: *** [src/console/dlt-convert] Error 1
CMakeFiles/Makefile2:295: recipe for target 'src/console/CMakeFiles/dlt-convert.dir/all' failed
make[1]: *** [src/console/CMakeFiles/dlt-convert.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2

I check and it turns out that this API dlt_check_envvar() is also used by dlt-user, so I suspect this change somehow affects other components.
The second thing is that as you could see from the error, it seems that there is something wrong in the source code when we are trying to change the API like that.
Please kindly check!
Regards

michael-methner:
Similar feature is implemented by https://github.com/COVESA/dlt-daemon/pull/421
and author is not reacting on questions -> closing this one.

PR #400, opened 2022-08-24
Feature/set host interface and allow multiple udp multicast addresses

alexanderirion:


src/console/dlt-receive.c
```diff
@@ -506,11 +514,30 @@ int main(int argc, char *argv[])
...
+
+        unsigned int servIPLength = 1; // Counting the terminating 0 byte
+        for (index = optind; index < argc; index++) {
+            servIPLength += strlen(argv[index]);
+            if (index > optind)
+            {
+                servIPLength++; // For the comma delimiter
+            }
+        }
+        if (servIPLength > 0) {
```
thanhbnq, line 526: `unsigned int servIPLength = 1; // Counting the terminating 0 byte`
so I believe this check is always true.
Please correct this.

michael-methner:
Was replaced by https://github.com/COVESA/dlt-daemon/pull/420

PR #401, opened 2022-08-29
systemd: add support for socket activation via systemd

alexmohr:
dlt-daemon now support activation via systemd socket.
This allows dlt-daemon to be started as soon as an application
is requesting access to the dlt socket.
This commits implements opening the socket only when
UNIX_SOCKET is used for IPC


michael-methner:
Looks good to me. Thank you very much. Verified with local environment

PR #402, opened 2022-08-29
logstorage: Truncate ECUid in Logstorage filter to prevent crash

andreirusu96:
This commit prevents a buffer overflow in case the supplied ECUID inside a Logstorage filter .conf file is larger than 4 bytes because it is currently not truncated anywhere else.


andreirusu96:
For example, the following Logstorage .conf file will cause dlt-daemon to crash:

[FILTER1]
LogAppName=TEST
ContextName=.*
...
EcuID=ECU1THISLINEWILLTRIGGERBUFFEROVERFLOW

With this fix, EcuID will be truncated to 4 bytes, as it is already done with AppID/ContextID.

PR #403, opened 2022-09-01
client: Fix Get Log Info response parse crash

andreirusu96:
Fix wrong cast in `dlt_getloginfo_conv_ascii_to_uint16_t` which would lead to a crash if the read value would be more than 255.
Fix missing `count_app_ids = 0` in `dlt_client_free_calloc_failed_get_log_info` which causes a crash if this method would be called, as the final de-init method will iterate again and access free'd memory.
Fix usage of `dlt_getloginfo_conv_ascii_to_id` which requires a bigger buffer because it also sets a NULL terminator.

### DIFF
--- a/src/lib/dlt_client.c
+++ b/src/lib/dlt_client.c
@@ -1146,6 +1146,7 @@ DLT_STATIC void dlt_client_free_calloc_failed_get_log_info(DltServiceGetLogInfoR
 
     free(resp->log_info_type.app_ids);
     resp->log_info_type.app_ids = NULL;
+    resp->log_info_type.count_app_ids = 0;
 
     return;
 }
@@ -1159,6 +1160,7 @@ DltReturnValue dlt_client_parse_get_log_info_resp_text(DltServiceGetLogInfoRespo
     int j = 0;
     char *rp = NULL;
     int rp_count = 0;
+    char temp_id[DLT_ID_SIZE + 1] = {0};
 
     if ((resp == NULL) || (resp_text == NULL))
         return DLT_RETURN_WRONG_PARAMETER;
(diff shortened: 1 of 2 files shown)


src/shared/dlt_common.c
```diff
@@ -4060,7 +4060,7 @@ int16_t dlt_getloginfo_conv_ascii_to_uint16_t(char *rp, int *rp_count)
     num_work[4] = 0;
     *rp_count += 6;
 
-    return (unsigned char)strtol(num_work, &endptr, 16);
+    return (unsigned short)strtol(num_work, &endptr, 16);
```
michael-methner, line 4063:
```suggestion
    return (uint16_t)strtol(num_work, &endptr, 16);
```

src/lib/dlt_client.c
```diff
@@ -1246,9 +1251,12 @@ DltReturnValue dlt_client_parse_get_log_info_resp_text(DltServiceGetLogInfoRespo
             /* get con id */
             dlt_getloginfo_conv_ascii_to_id(rp,
                                             &rp_count,
-                                            con->context_id,
+                                            temp_id,
```
michael-methner, line 1254: This change should not be required as dlt_getlonginfo_conv_ascii_to_id() takes care of the argument DLT_ID_SIZE. Please revert to original code.

andreirusu96: Hello! As described in the added code comment, `dlt_getlonginfo_conv_ascii_to_id()` does indeed have a length parameter, but the issue is that it also writes an **additional** NULL terminator after writing the specified `len` bytes. Since the `con->context_id` and `app->app_id` char arrays are of size **4**, the method will fill the buffer and will write another byte after the end of the ID array, **overwriting** one byte of the variable that resides next to those in their respective structures.

In other usages (App/Ctx descriptions), an additional byte is allocated specifically for the NULL terminator.

michael-methner:
The reported crash was fixed by #411 . Therefore closing the PR.

PR #405, opened 2022-09-05
Update dlt_for_developers.md

marques-bruno:
Fix error in suggested best-practice code snippet


michael-methner:
Hello @marques-bruno ,
the declaration of a variable in the for loop is allowed since C99.

michael-methner:
Hello @marques-bruno , thanks for the hint. that line was hidden in the diff.

PR #406, opened 2022-09-13
dlt_common.c: Change default logging_mode

wusto:
Up to now the default for logging is stdout (DLT_LOG_TO_CONSOLE).

Scenario:
program_a arg1 | dd of=/my-beauty-case

If program_a wants to log to dlt, but it fails to connect to the socket: socket /run/dlt-files/dlt cannot be opened, error:Resource temporarily unavailable. Retrying later...

The error will be logged by default to stdout leading in the best case to an error and in the worst to damaged partitions.

To solve this log to STDERR (DLT_LOG_TO_STDERR) by default.


thanhbnq:
Thank you very much for your PR @wusto

PR #411, opened 2022-10-05
Avoid memory corruption behind buffer wp in function dlt_getloginfo_conv_ascii_to_id

michael-methner:
- Introduced new function dlt_getloginfo_conv_ascii_to_string for '\0' terminated strings
- Avoid printing garbage characters in dlt-control after APID and CTID (which are not null terminated anymore)
- Added unit test for dlt_client_parse_get_log_info_resp_text and dlt_getloginfo_conv_ascii_to_string

### DIFF
--- a/include/dlt/dlt_common.h
+++ b/include/dlt/dlt_common.h
@@ -1637,7 +1637,7 @@ int16_t dlt_getloginfo_conv_ascii_to_int16_t(char *rp, int *rp_count);
  *
  * @param rp        char
  * @param rp_count  int
- * @param wp        char
+ * @param wp        char Array needs to be 1 byte larger than len to store '\0'
  * @param len       int
  */
 void dlt_getloginfo_conv_ascii_to_string(char *rp, int *rp_count, char *wp, int len);
@@ -1648,7 +1648,7 @@ void dlt_getloginfo_conv_ascii_to_string(char *rp, int *rp_count, char *wp, int 
  *
  * @param rp        char
  * @param rp_count  int
- * @param wp        char Array needs to be 1 byte larger than len to store '\0'
+ * @param wp        char
  * @param len       int
  * @return position of last read character in wp
  */
--- a/include/dlt/dlt_common.h
+++ b/include/dlt/dlt_common.h
@@ -1631,15 +1631,28 @@ int16_t dlt_getloginfo_conv_ascii_to_uint16_t(char *rp, int *rp_count);
  */
 int16_t dlt_getloginfo_conv_ascii_to_int16_t(char *rp, int *rp_count);
 
+
 /**
- * Convert get log info from ASCII to ID
+ * Convert get log info from ASCII to string (with '\0' termination)
  *
  * @param rp        char
  * @param rp_count  int
  * @param wp        char
  * @param len       int
  */
-void dlt_getloginfo_conv_ascii_to_id(char *rp, int *rp_count, char *wp, int len);
+void dlt_getloginfo_conv_ascii_to_string(char *rp, int *rp_count, char *wp, int len);
+
+
+/**
+ * Convert get log info from ASCII to ID (without '\0' termination)
+ *
+ * @param rp        char
+ * @param rp_count  int
+ * @param wp        char Array needs to be 1 byte larger than len to store '\0'
+ * @param len       int
+ * @return position of last read character in wp
+ */
+int dlt_getloginfo_conv_ascii_to_id(char *rp, int *rp_count, char *wp, int len);
 
 /**
  * Convert from hex ASCII to binary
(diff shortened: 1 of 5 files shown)


include/dlt/dlt_common.h
```diff
@@ -1631,15 +1631,28 @@ int16_t dlt_getloginfo_conv_ascii_to_uint16_t(char *rp, int *rp_count);
...
-void dlt_getloginfo_conv_ascii_to_id(char *rp, int *rp_count, char *wp, int len);
+void dlt_getloginfo_conv_ascii_to_string(char *rp, int *rp_count, char *wp, int len);
+
+
+/**
+ * Convert get log info from ASCII to ID (without '\0' termination)
+ *
+ * @param rp        char
+ * @param rp_count  int
+ * @param wp        char Array needs to be 1 byte larger than len to store '\0'
```
thanhbnq, line 1651: I think this comment should belong to function _to_string()

tests/gtest_dlt_common.cpp
```diff
@@ -4260,7 +4262,98 @@ TEST(dlt_get_minor_version, nullpointer)
...
+    DltServiceGetLogInfoResponse * resp = (DltServiceGetLogInfoResponse *)malloc(sizeof(DltServiceGetLogInfoResponse ));
+    DltReturnValue ret = (DltReturnValue) dlt_set_loginfo_parse_service_id(
+            input, &resp->service_id, &resp->status);
+    EXPECT_EQ(DLT_RETURN_OK, ret);
+    EXPECT_EQ(DLT_SERVICE_ID_GET_LOG_INFO, resp->service_id);
+    EXPECT_EQ(GET_LOG_INFO_STATUS_MAX, resp->status);
+
+    ret = dlt_client_parse_get_log_info_resp_text(resp, input);
+    EXPECT_EQ(DLT_RETURN_OK, ret);
+    EXPECT_EQ(DLT_RETURN_OK, ret);
```
thanhbnq, line 4287: I think this is a duplicated line

thanhbnq:
Hello @michael-methner ,
I also would like to separate the PR to 2 commits: one for UT and one for source code.

include/dlt/dlt_common.h
michael-methner: Thanks for the hint. I have fixed that.

tests/gtest_dlt_common.cpp
michael-methner: Thanks for the hint. I have removed the second line

michael-methner:
> Hello @michael-methner , I also would like to separate the PR to 2 commits: one for UT and one for source code.

Hello @thanhbnq , I want to keep them in one PR as only then the updated gtests are executed already by GitHub actions. Separating it into commits also does not make sense as they will be squashed anyhow.

andreirusu96:
Hello, @michael-methner ! Thanks for letting me know of your PR!

The code looks good! I just have a small remark: since you've introduced `dlt_getloginfo_conv_ascii_to_string()`, maybe you would like to also make use of it for app/ctx description parsing in [here](https://github.com/COVESA/dlt-daemon/blob/e961a8f7c0f655a4379c8c6bcf745f89604c453f/src/lib/dlt_client.c#L1335) and [here](https://github.com/COVESA/dlt-daemon/blob/e961a8f7c0f655a4379c8c6bcf745f89604c453f/src/lib/dlt_client.c#L1355).

Also, my PR ( #403 ) has 2 more improvements and you also suggested changing the cast in `dlt_getloginfo_conv_ascii_to_uint16_t()` to the same value as the return value.

Do you want to also include those improvements in your PR or would you prefer to do them in my PR (after a rebase)?

Thanks!

michael-methner:
> The code looks good! I just have a small remark: since you've introduced `dlt_getloginfo_conv_ascii_to_string()`, maybe you would like to also make use of it for app/ctx description parsing in [here](https://github.com/COVESA/dlt-daemon/blob/e961a8f7c0f655a4379c8c6bcf745f89604c453f/src/lib/dlt_client.c#L1335) and [here](https://github.com/COVESA/dlt-daemon/blob/e961a8f7c0f655a4379c8c6bcf745f89604c453f/src/lib/dlt_client.c#L1355).
> 
Thanks for the hint, actually this was my intention to use it here but somehow I missed to commit it. I fixed that one.

> Also, my PR ( #403 ) has 2 more improvements and you also suggested changing the cast in `dlt_getloginfo_conv_ascii_to_uint16_t()` to the same value as the return value.
> 
> Do you want to also include those improvements in your PR or would you prefer to do them in my PR (after a rebase)?
> 
> Thanks!

I would say we can merge this one now and we will add the casts with your PR after rebase. So you will be the author of that change.

thanhbnq:
Thank you @michael-methner  and @andreirusu96  for PR and discussion.
I merged PR and please continue your next PR :)

Happy coding!
Thanh

PR #415, opened 2022-10-12
test-pr:fix bus error for ./dlt-receive -a localhost

magiccaichm:
issue: fix bus error for ./dlt-receive -a localhost
analyze: at some times, we found some bus error in execute ./dlt-receive -a localhost, after i analyze , i found the address of "src" is odd number, in some system it will cause bus error.
fix: in some systems, the address shall follow byte alignment rule,so i changed the value-copy mechanism, please help review it.


magiccaichm:
code upload wrong

PR #420, opened 2022-10-25
dlt-receive: set host interface and allow multiple udp multicast addresses

thanhbnq:
This PR is working as revised version of #400 .
- Keep the committer name and email.
- Improve coding style


michael-methner:
Looks good to me, thank you very much.

PR #422, opened 2022-11-03
client: Fix Get Log Info response conversion method

andreirusu96:
Fix wrong cast in `dlt_getloginfo_conv_ascii_to_uint16_t` which would lead to a crash if the read value would be more than 255.
Fix missing `count_app_ids = 0` in` dlt_client_free_calloc_failed_get_log_info` which causes a crash if this method would be called, as the final de-init method will iterate again and access free'd memory.


michael-methner:
Hello @andreirusu96 ,
thanks for the patch.

PR #429, opened 2022-12-14
gateway: Fix Node handling and ECUid checks

andreirusu96:
Reset Receiver at connection close to prevent corruption: a partial log may be received from a Node, but in case there is an ECUid mismatch, the socket will be closed and that log will never be completed corrupting the receiver.
Prevent out of bounds reads
Fix connection select when processing passive node messages: fd's can use the same value for a new connection if a previous connection is closed so we need to also select based on connection status


lvklevankhanh (approved):
Look good to me.

PR #433, opened 2023-01-10
Fix atomic build on riscv64

LocutusOfBorg:
Add check for atomic need (e.g. this fixes the build on riscv64 architecture)


michael-methner:
Hello @LocutusOfBorg ,
thank you for providing the fix and documenting the license correctly.
But I rather do not pull in this commit as DLT is licensed under MPL license and I do not want to contaminate it with other licenses.

Which compilers do require this check? std::atomic is part of C++11, so actually a quite old feature.

LocutusOfBorg:
Its riscv64 architecture, some atomic are not optimizable.
Changing atomic<bool> to atomic<int> works in avoiding the whole -latomic linkage.

LocutusOfBorg:
Hello, this is fixed in a better way from 2224cddf140c060d92455ad3ee585e3abfc38eb2

michael-methner:
> Hello, this is fixed in a better way from [2224cdd](https://github.com/COVESA/dlt-daemon/commit/2224cddf140c060d92455ad3ee585e3abfc38eb2)

Do I understand you correctly, that the linked patch is also solving your issues on riscv64? Then I would close this PR.

LocutusOfBorg:
yes thanks!

PR #434, opened 2023-01-17
dlt-convert: Fix memory leak by calling dlt_file_free

lvklevankhanh:
Free the index which already allocated memory in dlt_file_read phase

Signed-off-by: Le Van Khanh <Khanh.LeVan@vn.bosch.com>


zj3t:
Checked the contents of the patch.

PR #435, opened 2023-01-17
dlt-user: fix potential non closed socket in init/free

alexmohr:
wait until the house keeper is up and running
this makes sure dlt_init / dlt_free in close succession closes this thread properly.
Otherwise we might keep the dlt socket open
Check if the thread is running every 10ms

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -111,6 +111,10 @@ static char dlt_daemon_fifo[DLT_PATH_MAX];
 static sem_t dlt_mutex;
 static pthread_t dlt_housekeeperthread_handle;
 
+/* Sync housekeeper thread start */
+pthread_mutex_t dlt_housekeeper_running_mutex = PTHREAD_MUTEX_INITIALIZER;
+pthread_cond_t dlt_housekeeper_running_cond = PTHREAD_COND_INITIALIZER;
+
 /* calling dlt_user_atexit_handler() second time fails with error message */
 static int atexit_registered = 0;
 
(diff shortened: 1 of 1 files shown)


src/lib/dlt_user.c
```diff
@@ -3736,10 +3736,12 @@ static void dlt_user_cleanup_handler(void *arg)
     DLT_SEM_FREE();
 }
 
-void dlt_user_housekeeperthread_function(__attribute__((unused)) void *ptr)
+void dlt_user_housekeeperthread_function(void *ptr)
 {
     struct timespec ts;
     bool in_loop = true;
+    bool *thread_running = (bool*)ptr;
+    *thread_running = true;
```
michael-methner, line 3744: Hello @alexmohr ,
in which platform have you faced this issue?
I guess the assignment to true needs to be shifted below pthread_sigmask (for Android) and calling pthread_cleanup_push(). Otherwise a race condition still exists.

src/lib/dlt_user.c
```diff
@@ -4945,14 +4947,25 @@ void dlt_user_test_corrupt_message_size(int enable, int16_t size)
...
     }
 
+     /* wait until the house keeper is up and running
+     * this makes sure dlt_init / dlt_free in close succession
+     * closes this thread properly.
+     * Otherwise we might keep the dlt socket open
+     * Check if the thread is running every 10ms
+     * */
+    while (!housekeeper_thread_running) {
+        usleep(10000);
```
michael-methner, line 4966: Can you use a semaphore instead? looping and sleeping is not that nice.

src/lib/dlt_user.c
alexmohr: Issue occurs on linux but only in unit tests when running them on a very fast machine. I moved the call and replaced the sleep with a condition variable

michael-methner: Looks good to me. Thank you very much.

PR #437, opened 2023-01-27
Check for negative index in dlt_file_message

### LINKED ISSUE #436: [dlt-sortbytimestamp crash] Unreadable memory reference in my opinion
Hi COVESA Team!
Contact us to report vulnerabilities.

## Summary
I done dlt-sortbytimestamp(+dlt-* binary)  source code auditing . 
And found that the exception handling was insufficient.
Since I'm not familiar with the dlt structure, I created a simple fuzzer to fuzzing the dlt-sortbytimestamp binary.

And I found a crash, and added ASAN (Address Sanitizer) to get additional information.

## Bug Log
image

image

In my opinion, when the pointer is moved in the fseek() function, the crash occurs because it is moved to a place that cannot be referenced.
Or, Maybe it's caused by size tampering in the header of the DLT file.
But I don't know the detailed information of the DLT file.

image
[crash.dlt]

image
[output.dlt]

Looking at the output file, it was confirmed that it terminated in the middle.

## Files
I will attach it.
[covesa.zip](https://github.com/COVESA/dlt-daemon/files/10487695/covesa.zip)

It may not be a big deal, it may be a bug (vulnerability), but I am reporting it because I hope you guys will fix it.
Thanks :)

michael-methner:
Fixes #436

### DIFF
--- a/src/shared/dlt_common.c
+++ b/src/shared/dlt_common.c
@@ -1718,7 +1718,7 @@ DltReturnValue dlt_file_message(DltFile *file, int index, int verbose)
         return DLT_RETURN_WRONG_PARAMETER;
 
     /* check if message is in range */
-    if (index >= file->counter) {
+    if (index < 0 || index >= file->counter) {
         dlt_vlog(LOG_WARNING, "Message %d out of range!\r\n", index);
         return DLT_RETURN_WRONG_PARAMETER;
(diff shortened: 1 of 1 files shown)


lvklevankhanh (approved):
LGTM

PR #438, opened 2023-01-31
house-keeper: remove infinite wait

alexmohr:
in case the house keeper thread does not start
we should not wait indefinitely but abort
the wait after a 10s

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -113,7 +113,7 @@ static pthread_t dlt_housekeeperthread_handle;
 
 /* Sync housekeeper thread start */
 pthread_mutex_t dlt_housekeeper_running_mutex = PTHREAD_MUTEX_INITIALIZER;
-pthread_cond_t dlt_housekeeper_running_cond = PTHREAD_COND_INITIALIZER;
+pthread_cond_t dlt_housekeeper_running_cond;
 
 /* calling dlt_user_atexit_handler() second time fails with error message */
 static int atexit_registered = 0;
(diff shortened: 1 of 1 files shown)


alexmohr:
fyi @methner this is an addition to https://github.com/COVESA/dlt-daemon/pull/435

src/lib/dlt_user.c
```diff
@@ -4972,22 +4972,21 @@ int dlt_start_threads()
...
-            &time_to_wait);
-        if (signal_status != 0) {
-            dlt_log(LOG_CRIT, "Failed to wait for house keeper thread!\n");
-        }
+    signal_status = pthread_cond_timedwait(
+        &dlt_housekeeper_running_cond, 
+        &dlt_housekeeper_running_mutex,
+        &time_to_wait);
+    if (signal_status != 0) {
+        dlt_log(LOG_CRIT, "Failed to wait for house keeper thread!\n");
```
michael-methner, line 4987: If signal_status != 0, the housekeeper thread is not running correctly. 
I would propose to call dlt_stop_threads() the thread and to return a "-1" here as well.

dlt_stop_threads() should also be called in case dlt_user_trace_network_sgmented_thread could not be started.

This ensures that we leave DLT does not keep any threads running in case it returns an error code.

alexmohr: good idea :+1: updated the commit

alexmohr: okay so I had to make some more changes, especially we have to take into account the scenario that the clock moves in while we're in this loop. 
To prevent issues with that CLOCK_MONONTIC_RAW is used. 
Reasoning for the implementation and why I'm using the loop again is given in the code.

alexmohr: oh and one more edit... I'm using `CLOCK_MONOTONIC` instead of `CLOCK_MONONTIC_RAW` because `RAW` is not supported by some variants of glibc and it will return `EINVAL` or even call `abort` in some circumstances. But only if it actually has to wait on the cond var, that's why I didn't catch it in the first version but only after some extensive testing.

lti9hc:
I have tested PR on my local environment, There is no impact, so I think we can merge this PR now

PR #441, opened 2023-02-09
Fix memory leak

lvklevankhanh:
Free the ecuid_conf in case of memory alllocated

Signed-off-by: Le Van Khanh <Khanh.LeVan@vn.bosch.com>


zj3t:
@lvklevankhanh 
It's really good. The patch worked fine.

PR #445, opened 2023-02-27
dlt-logd-converter: fixes android 12 compilation

sebastienraillet:
# Summary
This PR is about fixing a compilation issue which appear starting on android 12 for android-logd-converter

# Description
* Since android 12 release (and the introduction of this [commit](https://chromium.googlesource.com/aosp/platform/system/logging/+/b674866203f05957b2ac5db94c3c0fe3d1d36793)), the macro ANDROID_LOG_RDONLY and friends have been removed (see the commit for the explanation). Therefore, dlt-logd-converter can't compile anymore since this version.
* The modification have been made in such way to allow dlt-logd-converter to compile also on version prior to android 12

### DIFF
--- a/src/android/dlt-logd-converter.cpp
+++ b/src/android/dlt-logd-converter.cpp
@@ -22,7 +22,6 @@
 #include <csignal>
 
 #include <log/log_read.h>
-#include <log/logprint.h>
 
 #include <dlt.h>
 
@@ -51,7 +50,7 @@ static inline struct logger *init_logger(struct logger_list *logger_list, log_id
 static struct logger_list *init_logger_list(bool skip_binary_buffers)
 {
     struct logger_list *logger_list;
-    logger_list = android_logger_list_alloc(ANDROID_LOG_RDONLY, 0, 0);
+    logger_list = android_logger_list_alloc(O_RDONLY, 0, 0);
     if (logger_list == nullptr) {
         DLT_LOG(dlt_ctx_self, DLT_LOG_FATAL, DLT_STRING("could not allocate logger list"));
         return nullptr;
--- a/Android.bp
+++ b/Android.bp
@@ -194,9 +194,6 @@ cc_binary {
         "libdlt",
         "liblog",
     ],
-    include_dirs: [
-        "system/core/include",
-    ],
 }
 
 // vim: ft=python


src/android/dlt-logd-converter.cpp
```diff
@@ -50,8 +54,13 @@ static inline struct logger *init_logger(struct logger_list *logger_list, log_id
 
 static struct logger_list *init_logger_list(bool skip_binary_buffers)
 {
+#if __ANDROID_API__ >= 31
+    int mode = 0;
```
michael-methner, line 58: Hello @sebastienraillet ,
can you please use "O_RDONLY" instead of the hardcoded "0"?

sebastienraillet: Hello @michael-methner,
Yes sure 👍  In fact, I didn't see that previously, **ANDROID_LOG_RDONLY** was a define on **O_RDONLY**.
We can even directly pass O_RDONLY in both case (before and after API 31) avoiding the #if / #endif.
Which solution do you prefer ?

minminlittleshrimp: Hello @sebastienraillet 

For Android S and above, the path change: https://cs.android.com/android/platform/superproject/+/master:system/logging/liblog/include/log/
Hence please correct Android blue print file: 
https://github.com/COVESA/dlt-daemon/blob/master/Android.bp#L198

For backward compatibility, I still have no idea. If anybody has good suggestions please comment!

src/android/dlt-logd-converter.cpp
```diff
@@ -21,8 +21,12 @@
 #include <cstdlib>
 #include <csignal>
 
+#if __ANDROID_API__ >= 31
+#include <log/log.h>
```
minminlittleshrimp, line 25: No need I think.
Since the current implementation does not require all API.

minminlittleshrimp:
Hello @sebastienraillet 

For Android S and above, the path change: https://cs.android.com/android/platform/superproject/+/master:system/logging/liblog/include/log/
Hence please correct Android blue print file: 
https://github.com/COVESA/dlt-daemon/blob/master/Android.bp#L198

For backward compatibility, I still have no idea. If anybody has good suggestions please comment!

src/android/dlt-logd-converter.cpp
sebastienraillet: Hello @minminlittleshrimp,

At the beginning, this was the compilation issue I had : 
`external/dlt-daemon/src/android/dlt-logd-converter.cpp:31:10: fatal error: 'log/logprint.h' file not found`

As the liblog [README.md](https://cs.android.com/android/platform/superproject/+/master:system/logging/liblog/;l=1?q=liblog&sq=&ss=android%2Fplatform%2Fsuperproject) is giving as an example the include of _log.h_ directly, I though it was the right way to do for Android 12 and above. This was the intention of my commit.

However, after some investigation I discovered that : 
* The log/logprint.h include is useless (we aren't using any API from this header file). Everything is compiling properly without it
* This header **shouldn't** be included by vendor application. Your comment about making a change inside the .bp put me on the right track : 
> Hello @sebastienraillet
> 
> For Android S and above, the path change: https://cs.android.com/android/platform/superproject/+/master:system/logging/liblog/include/log/ Hence please correct Android blue print file: https://github.com/COVESA/dlt-daemon/blob/master/Android.bp#L198
> 
> For backward compatibility, I still have no idea. If anybody has good suggestions please comment!

In fact, if you have a look to liblog Android.bp, you will see a [special management](https://cs.android.com/android/platform/superproject/+/android-12.0.0_r34:system/logging/liblog/Android.bp;drc=b2cb35cc4d5da1840c8db4c133e49a551c5c694f;l=77) for vendor use case. A vendor application will only see the headers available inside the [include_vndk ](https://cs.android.com/android/platform/superproject/+/android-12.0.0_r34:system/logging/liblog/include_vndk/) folder (and as you can imagine, logprint.h isn't part of this 😉)

The line [198](https://github.com/COVESA/dlt-daemon/blob/master/Android.bp#L198) inside the Android.bp should be removed. Android build system will automatically add liblog headers path as we are linked to it through _shared_libs_ keyword.

I don't know if my explanation is clear. If not, please don't hesitate to tell me 👍 

To summarize and if you agree we have 2 choices : 
* I will redo my commit by only removing _logprint.h_ include directive as well as #if / #endif management which become useless.
* We include only _log.h_ directly. For me, it sounds a reasonnable solution based on [this comment](https://cs.android.com/android/platform/superproject/+/android-12.0.0_r34:system/logging/liblog/include_vndk/log/log.h;drc=1ef5858d728fc33a1510522b276b7a868af8e1c7;l=1) inside the log.h file. It will ensure that, even if Google is reworking headers behind log.h, we won't be affected

We should also remove the line [198](https://github.com/COVESA/dlt-daemon/blob/master/Android.bp#L198) inside Android.bp because it isn't the right way to do to include headers from an another library. Android build system will do this automatically for us as we are linked to liblog

sebastienraillet:
Hello @minminlittleshrimp,
See my comment [here](https://github.com/COVESA/dlt-daemon/pull/445#discussion_r1134024553) for full explanation 😃

src/android/dlt-logd-converter.cpp
sebastienraillet: Hello @minminlittleshrimp,

> I am not an Android expert, but there's a first time for everything

No problem 😉 AOSP build system isn't so easy at the first approach.

I agree on point 1. and 2.

About 3. point, why not using _O_RDONLY_ in both cases ?
Before android 12, ANDROID_LOG_RDONLY was simply a define on O_RDONLY. Thefore, make sense for me to use O_RDONLY and this will make the code compatible for all versions without the need to manage #if / #endif. Easier to read and maintain of my point of view.

What do you think ? 

Regards,

minminlittleshrimp: Great, I have not noticed that.
I agree with you on point 3, please kindly go ahead with the change.

Thank you

sebastienraillet: I've pushed the branch including the modification discussed above. We should be all aligned now 👍

michael-methner (approved):

PR #447, opened 2023-03-01
When MaintainLogstorageLogLevel=OFF, we should not change log level o…

jianjun-huang:
…f any apps/contexts which are connected to this daemon directly as well as are passively connected to this daemon

### DIFF
--- a/src/daemon/dlt_daemon_offline_logstorage.c
+++ b/src/daemon/dlt_daemon_offline_logstorage.c
@@ -893,6 +893,9 @@ void dlt_daemon_logstorage_update_application_loglevel(DltDaemon *daemon,
 
     handle = &(daemon->storage_handle[dev_num]);
 
+    if(handle->maintain_logstorage_loglevel == 0)
+        return;
+
     if ((handle->connection_type != DLT_OFFLINE_LOGSTORAGE_DEVICE_CONNECTED) ||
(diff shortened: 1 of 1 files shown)


michael-methner:
Hello @jianjun-huang ,
this is not as per the documentation at: https://github.com/COVESA/dlt-daemon/blob/master/doc/dlt_offline_logstorage.md#maintain-logstorage-log-level-implementation

jianjun-huang:
Hi Michael:

Thank you for your review and response.

We still have some questions about this.
By setting MaintainLogstorageLogLevel=OFF or MaintainLogstorageLogLevel=0, the clients are able to update any log level to user contexts.


We agree about this design, but when reviewing the code in dlt_daemon_logstorage_update_application_loglevel,  we will find that logstorage is trying to use configured log level to replace user defined log level.

The test steps in gateway mode are :

  1.  In log storage configuration, the log level is VERBOSE,
  2.  The user changed log level to FATAL via dlt-viewer.
  3.  The user tried to get log info in dlt-viewer several times, then found that log level was rolled back to VERBOSE. (The getLogInfo request from Dlt-Viewer will trigger calling dlt_daemon_logstorage_update_application_loglevel)

If MaintainLogstorageLogLevel =OFF, why do we need to call dlt_daemon_logstorage_update_application_loglevel?

PR #463, opened 2023-04-05
dlt-user: fix crash with certain strings

alexmohr:
make sure that a string with exactly UINT16_MAX bytes does not overflow the dlt buffer

Given the example below there is a crash is libdlt 
```c++
#include <dlt/dlt.h>
#include <cstring>
#include <string>

DLT_DECLARE_CONTEXT(mycontext)

int main()
{
    DLT_REGISTER_APP("FOOB", "TEST DEMO");
    DLT_REGISTER_CONTEXT(mycontext, "CTX1", "Test Context for Logging");
    size_t len = 65535;
    char* txt = (char*)malloc(len);
    memset(txt, 'c', len);
    
    // will crash
    DLT_LOG(mycontext, DLT_LOG_INFO, DLT_STRING(txt));
}
```

The expansion of the macro looks like this
```c
do {
    DltContextData log_local;
    int dlt_local;
    dlt_local = dlt_user_log_write_start(&mycontext, &log_local, DLT_LOG_INFO);
    if (dlt_local == DLT_RETURN_TRUE) {
        (void)dlt_user_log_write_string(&log_local, txt);
        (void)dlt_user_log_write_finish(&log_local);
    }
} while (0
```

The crash now happens due to the following chain of events.
`dlt_user_log_write_start(&mycontext, &log_local, DLT_LOG_INFO);` is calling `dlt_user_log_write_string_utils_attr`
Which is casting the string length into `uint16_t`. This is fine as our string is exactly as long as `uint16_t` allows.
Now `dlt_user_log_write_sized_string_utils_attr` is called.
This sets `uint16_t arg_size = (uint16_t) (length + 1)`. Which will overflow the uint16_t and `arg_size` will be `0`.

```c
// new_log_size = 2
// will be increment to 6 if verbose mode is on
size_t new_log_size = log->size + arg_size + sizeof(uint16_t);
```

The code above also won't detect that the string is too large as the uin16_t also overflows and the sum will be much smaller as `dlt_user.log_buf_len` resulting in the truncation check to eval to false.

```c
/* Check log size condition */
if (new_log_size > dlt_user.log_buf_len)
// no jump into if body as new_log_size is small enough to fit into the buffer
// thus no truncation
```

Now `log->size` is incremented in several places. 
In my example with enabled verbose mode it's `6`

```c
 /* Whole string will be copied */
memcpy(log->buffer + log->size, text, length);
```

The code above is now copying into the buffer at `buffer + log->size`. 
As the original length with 65535 is used we're writing 6 bytes behind the actual buffer.
With larger strings, their length is still a multiple of `uint16_t max`, the same crash occurred.
If it's not a multiple the application does not crash but reads a certain amount of bytes from the buffer, truncating the log without any message that it was cut of.

It's fixed by using size_t for the calculation, so the truncation is working properly.

After the fix ` case DLT_RETURN_USER_BUFFER_FULL:` is used and the truncated string will be logged.

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -221,7 +221,7 @@ static void dlt_user_trace_network_segmented_thread_segmenter(s_segmented_data *
 #endif
 
 static DltReturnValue dlt_user_log_write_string_utils_attr(DltContextData *log, const char *text, const enum StringType type, const char *name, bool with_var_info);
-static DltReturnValue dlt_user_log_write_sized_string_utils_attr(DltContextData *log, const char *text, uint16_t length, const enum StringType type, const char *name, bool with_var_info);
+static DltReturnValue dlt_user_log_write_sized_string_utils_attr(DltContextData *log, const char *text, size_t length, const enum StringType type, const char *name, bool with_var_info);
 
 
 static DltReturnValue dlt_unregister_app_util(bool force_sending_messages);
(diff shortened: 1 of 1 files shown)


alexmohr:
Another way to test the faulty version is by running dlt-example-user
```
 dlt-example-user -n 1 $(cat /dev/urandom | tr -dc '[:alpha:]' | fold -w ${1:-65535} | head -n 1)
```

michael-methner (approved):
Looks good to me. Local verification successful

PR #467, opened 2023-04-17
Update status badges on README.md

michael-methner:
Removed deprecated LGTM, added CodeQL


minminlittleshrimp:
Hello @michael-methner 

Thank you, then for the code coverage, should we give it a try for gcov?
We already make use of them before and we could try to implement it into github workflow.

Regards

lvklevankhanh (approved):

PR #470, opened 2023-04-24
watchdog: improve dlt watchdog

alexmohr:
* watchdog can be configured to notify systemd only if messages are received
* activate this via:
  * -DWITH_SYSTEMD_WATCHDOG_ENFORCE_MSG_RX=ON
  * -DWITH_SYSTEMD_WATCHDOG_ENFORCE_MSG_RX_DLT_SYSTEM=ON

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,8 @@ endif()
 option(BUILD_SHARED_LIBS      "Set to OFF to build static libraries"                                             ON)
 option(WITH_SYSTEMD           "Set to ON to create unit files and systemd check on dlt-daemon startup"           OFF)
 option(WITH_SYSTEMD_WATCHDOG  "Set to ON to use the systemd watchdog in dlt-daemon"                              OFF)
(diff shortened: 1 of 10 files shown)


michael-methner:
Hello @alexmohr ,
thank you for the proposal, but I disagree with the feature.
In my point of view, the trigger of the watchdog should only indicated that dlt-deamon and dlt-system is still in normal operation. Relying in the receiving of log messages to trigger the watchdog may have unintended side effects. clients should instead implement a own watchdog mechanism and not rely on DLT for this.

So I tend to not merge this change. Sorry.

alexmohr:
I do agree that clients should not rely on this mechanism as watchdog. This change just makes sure that data of any application is still sent to daemon. 
In our project an absence of logs, even for a few seconds, means that something is wrong with the daemon or dlt-system, because a very large number of logs is processed each seconds. 
Absence of them does not mean an application is not working anymore but the system / daemon itself has an issue that prevents the processing of logs.
Therefore it improves that watchdog of the daemon and system in way that makes logs are actually processed / received and not some timer is still working.
The operation of the later does not at all indicate if the service is still working properly.

We do have different targets though, and it's disabled on some of them.

michael-methner (approved):

PR #472, opened 2023-04-24
Remove use of DLT_LOG in signal handler

michael-methner:
DLT_LOG is using internally function which are not async-signal safe (like calloc, vsnprintf). Therefore, it needs to be removed from the signal handler


lvklevankhanh (approved):
Look good to me.

PR #475, opened 2023-04-25
fix pedantic warning for compilers using C99 standard

alexmohr:
fix pedenantic warnings some compilers produce when using variadic macros

### DIFF
--- a/include/dlt/dlt_user_macros.h
+++ b/include/dlt/dlt_user_macros.h
@@ -53,6 +53,14 @@
 **  mk          Markus Klein               Fraunhofer ESK                     **
 *******************************************************************************/
 
+/**************************************************************************************************
++* Compilers may complain about not using variadic macros if they are supported
++**************************************************************************************************/
+#if __STDC_VERSION__ >= 199901L || __cplusplus >= 201103L
+/* DLT_ENABLE_C_VARIADIC_MACROS code */
+#define DLT_ENABLE_C_VARIADIC_MACROS
+#endif
+
 /*******************************************************************************
 **                      Revision Control History                              **
 *******************************************************************************/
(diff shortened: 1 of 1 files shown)


michael-methner:
Hello @alexmohr ,
thanks for the patch. I would like to avoid the code duplication. Anyhow, I tried to compile  DLT with C89 standard for fun but it failed due to various issues.
Would you like to rework the patch to only support C99? Then we can get rid of the duplicated macros.

alexmohr:
Hi, thanks for testing it. I upstreamed this patch, because it's part of our internal patches and with the update the 2.18.9 we're trying to upstream all of them.
I don't think it's worth the effort to make dlt compliant with C99. The patch has been made internally because the macro will be used in some components that are compiling with C99. 
I'm note sure with we proceed with this pull request or close it because it's no real benefit for upstream

michael-methner:
Hello @alexmohr ,
i experimented with the gcc compiler, but I could not find any standard which is compiling the named variadic args without an pedantic warning:
 warning: ISO C does not permit named variadic macros
 warning: ISO C++ does not permit named variadic macros
It tried following standards: 
C: c89 c99 c11 gnu99 gnu11
C++: c++03 c++11 c++14 gnu++98 gnu++03 gnu++11 gnu++14

As per my understanding the "named variadic args" where supported by GNU C++ only before it became a standard (as a gnu extension). Nowadays the compiler even generates a warning when used with old standards.

The anonymous variadic macros only generates a warnings on c89, c++03 and older:
warning: anonymous variadic macros were introduced in C99 [-Wvariadic-macros]
warning: anonymous variadic macros were introduced in C++11 [-Wvariadic-macros]

Therefore I do not see any benefit of merging this change.  I would close this PR if you agree.

alexmohr:
Thanks for giving it a try. It makes sense to close it :+1:

PR #476, opened 2023-04-25
dltuser: add method to configure ring buffer size per application.

alexmohr:
Logs of applications are kept in a ring buffer.
Each application (ie user) using logging functionality of dlt-daemon should be allowed to write
to ring buffer a limited asize of logs not to exceed limits available in terms of memory and performance. This patch is useful for upstream.
Because, first, the ability of limiting logs
of each application will make the use of system memory resources without exceeding the overall capacity limits. Secondly the logs of each pplication will be limited as a guard to excessive logging - ie logging in
for loops- which could not easily
be determined since most of them occur in extreme conditions.

### DIFF
--- a/include/dlt/dlt_user.h.in
+++ b/include/dlt/dlt_user.h.in
@@ -1231,6 +1231,13 @@ int dlt_user_atexit_blow_out_user_buffer(void);
  */
 DltReturnValue dlt_user_log_resend_buffer(void);
 
+ /**
+ * Try to set max size of ring buffer for dlt user
+ * @param size new size of ring buffer in bytes
+ * @return Value from DltReturnValue enum
+ */
+DltReturnValue dlt_user_set_max_ring_buffer_size(uint32_t size);
+
 /**
  * Checks the log level passed by the log function if enabled for that context or not.
  * This function can be called by applications before generating their logs.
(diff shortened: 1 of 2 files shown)


lvklevankhanh:
Hi @alexmohr ,

Changing the code is fine with me. However, we need to verify in the local environment with the following points:
- What happens if "size" is smaller than current size of the buffer?
- Is it allowed to call the function after buffer was created?
- How to test this PR?
Could you please share us some information if you have verified these points?
Thank you!

michael-methner:
Hello @alexmohr ,
is there a specific reason why you do not use the environment variable DLT_USER_BUFFER_MAX for this? Do you need this to be changed dynamically during runtime of the application?

I feel that we might run into a race condition when this function is called as there is the buffer already allocated. So at least buf->size has to be checked as well before setting the value.

michael-methner:
Hello @alexmohr ,
how shall we continue with this PR?

alexmohr:
Hi @michael-methner , the PR as is should not be merged as it has some caveats like reducing the size of the buffer is not possible. I'm going to close it for now, as we want to get rid of it internally anyways. 
If we can't get rid of it we will rework it and create a new pull request for.

PR #479, opened 2023-04-27
build: add static lib only if necessary

alexmohr:
adding static library should only be enabled
if the dlt library is build static.
Adding this to the automotive-dlt.pc.in
when static libraries are disabled
yocto build will break


michael-methner:
Thanks for the contribution. Looks good to me.

PR #484, opened 2023-05-04
gtest: Update googletest installation process

minminlittleshrimp:
gtest: Update googletest installation process
Since googletest is external project, it should only be installed and built in need.
This patch tries to perform the download and unpacking steps at configure time.
Besides, the gtest 1.7 is quite old and misses some macros there when testing in Android, and the latest version of googletest does not support C++11 anymore but stop at v.1.11.0. Hence using this one.


minminlittleshrimp:
Update the latest googletest cmake config, handle googletest cmake installation at config time.

CMakeLists.txt
```diff
@@ -300,10 +300,31 @@ add_subdirectory(src)
...
+
+    # Add googletest directly to our build. This adds the following targets:
+    # gtest, gtest_main, gmock and gmock_main
+    add_subdirectory("${CMAKE_BINARY_DIR}/googletest-src"
+        "${CMAKE_BINARY_DIR}/googletest-build"
+    )
 
-    if (NOT GTEST_FOUND)
-        add_subdirectory( gtest-1.7.0 )
+    # The gtest/gmock targets carry header search path dependencies
```
minminlittleshrimp, line 321: Remove these lines 321-327 since dlt requires cmake > 3.3

michael-methner:
Hello @minminlittleshrimp ,
as discussed via phone, please use "git submodule" for the gtest instead of the cmake mechansim. Thank youl.

PR #490, opened 2023-05-25
dlt-qnx-system: prevent message loss in high load situations

alexmohr:
* Add wait method, which waits up to 5 seconds for buffer space to become available. This method tries to flush the buffer via tcp every 10ms. An error will be logged when there was not enough space. The error shows up in dlt to simplify debugging.
* improve cmake that more qnx compilers are detected
* log missing context mapping only once, this greatly reduces noise and the load of the system, as previously the message was logged with every log of an unmapped application

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,6 +82,7 @@ option(WITH_DLT_LOG_LEVEL_APP_CONFIG "Set to ON to enable default log levels bas
 
 set(DLT_IPC "FIFO" CACHE STRING "UNIX_SOCKET,FIFO")
 set(DLT_USER "genivi" CACHE STRING "Set user for process not run as root")
+set(DLT_QNX_SLOG_ADAPTER_WAIT_BUFFER_TIMEOUT_MS "100" CACHE STRING "Timeout in milliseconds to wait before messages are dropped when input buffer is full")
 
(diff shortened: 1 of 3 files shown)


michael-methner:
Hello @alexmohr ,
thanks for the patch.
I am totally fine to merge the two items:
- improve cmake that more qnx compilers are detected
- log missing context mapping only once, this greatly reduces noise and the load of the system, as previously the message was logged with every log of an unmapped application

I am having a little problem with the change to wait for 5 seconds.
1) How does slog behave when it is blocked for 5 seconds?
2) It would be good to have this behavior configurable
I assume that this situation most likely happens during startup for early messages which are sent while dlt is not yet started.

alexmohr:
>     1. How does slog behave when it is blocked for 5 seconds?
In our project there's a of stuff logged into slog, especially during startup. We never noticed any issues with this. Unfortunately the qnx manual does not give any further hints on how fast this callback should be done. 
Most likely the message are received via qnx message passing and thus should have no influence on the performance of slog itself.  

>     2. It would be good to have this behavior configurable
>        I assume that this situation most likely happens during startup for early messages which are sent while dlt is not yet started.

I made it configurable via cmake and changed to the default to 100ms.
The most relevant time is when the the dlt-qnx-system has just started and is receiving all messages from the slog since startup. At this point dlt-daemon is likely to be up already.

michael-methner:
Hello @alexmohr ,
thanks for the quick rework.

PR #511, opened 2023-07-27
Switch from GENIVI to COVESA

minminlittleshrimp:
GENIVI should be changed to COVESA:

+ Change config, name path, userid from genivi to covesa
+ Change GENIVI in urls, src code descriptions, etc to COVESA
+ Change logo from GENIVI to COVESA (remove logo image and replace with direct url from COVESA webpage)


lvklevankhanh (approved):
Look good to me.

PR #513, opened 2023-08-02
Removed null-terminated from payload

LukeTheEngineer:
On page 53 of the Autosar FO R22-11, the specification states the data payload of type String(STRG) must be constructed without a special terminator item such as '\0' 

@mbehr1 pointed out that since the length of the string is already determined, the need for a null terminator is useless in that context. 

**Before**
`case DLT_RETURN_OK: 
{ /* Whole string will be copied */
     memcpy(log->buffer + log->size, text, length); 
    **The input string might not be null-terminated, so we're doing that by ourselves** 
    log->buffer[log->size + length] = '\0'; 
    log->size += arg_size; 
    break; 
}`

**After** 
`    {
        /* Whole string will be copied */
        memcpy(log->buffer + log->size, text, length);

        /* The input string might not be null-terminated, so we're doing that by ourselves */

        /*The length of the string is already determined. Why need the null-terminator???*/
        //log->buffer[log->size + length] = '\0';   //*[Deprecated]

        log->size += arg_size;
        break;
    }`

I would like to know your thoughts on why this was added and if there was any specific meaning to this.

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -2707,7 +2707,7 @@ static DltReturnValue dlt_user_log_write_sized_string_utils_attr(DltContextData 
         memcpy(log->buffer + log->size, &type_info, sizeof(uint32_t));
         log->size += sizeof(uint32_t);
     }
-
+    
     memcpy(log->buffer + log->size, &arg_size, sizeof(uint16_t));
     log->size += sizeof(uint16_t);
 
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -2733,8 +2733,12 @@ static DltReturnValue dlt_user_log_write_sized_string_utils_attr(DltContextData 
     {
         /* Whole string will be copied */
         memcpy(log->buffer + log->size, text, length);
+
         /* The input string might not be null-terminated, so we're doing that by ourselves */
-        log->buffer[log->size + length] = '\0';
+        
+        /*The length of the string is already determined. Why need the null-terminator???*/
+        //log->buffer[log->size + length] = '\0';   //*[Deprecated]
+
         log->size += arg_size;
         break;
     }


src/lib/dlt_user.c
```diff
@@ -2733,8 +2733,12 @@ static DltReturnValue dlt_user_log_write_sized_string_utils_attr(DltContextData
     {
         /* Whole string will be copied */
         memcpy(log->buffer + log->size, text, length);
+
         /* The input string might not be null-terminated, so we're doing that by ourselves */
-        log->buffer[log->size + length] = '\0';
+        
+        /*The length of the string is already determined. Why need the null-terminator???*/
+        //log->buffer[log->size + length] = '\0';   //*[Deprecated]
```
mbehr1, line 2740: I think that's not enough. line 2621 adds the null term as +1 to the strlen. That needs to be removed as well (at least)

LukeTheEngineer (commented):
How much could the null terminators affect the user? Also, is this the only file the null-terminators are present in?

src/lib/dlt_user.c
michael-methner: Hello @LukeTheEngineer ,
yes, to be according to autosar specification, both places needs to be fixed otherwise long messages which will truncated will carry still NULL termination.

I fixed this at my local setup but noticed that a lot of the unit tests are failing then but I do not have the time to investigate this further now. 

We had in the past some problems with NULL-termination of strings. So I am a little bit hesitant to merge such a fix.  Even if it is according to spec, a lot of legacy code may nowadays rely on this NULL-termination (like the DLT unit tests). As I cannot think of any issues arising from this additional NULL-termination, I would prefer to not merging this but I am open to arguments here.


@mbehr1 , @LukeTheEngineer , @DLT-community: What's your opinion on this?

mbehr1: @michael-methner I stumbled exactly upon this problem on "relying" on null termination. (see https://github.com/mbehr1/adlt/commit/a17c59f5c69b7a7ab362ac026a0a95ef00f788e5 )
But the problem is that you cannot rely on this as there are some tools that create strings non null-terminated. So I'd better change this to always "non-null" / according to spec. than to rely on something which is not granted by spec and which not all tools do.

LukeTheEngineer: @michael-methner How frequently does NULL termination occur in the entire code base that affects this specific specification?

Perhaps there could be a way to perform the job of a NULL terminator without explicitly adding them? Then again, I am still relatively new to dlt-daemon, and still trying to understand how everything works.

michael-methner: @mbehr1: if you are consumer of these strings, you cannot rely on the null-termination. If you are a producer of the strings, I think an additional null-termination does not hurt. As standard-conform consumers will handle both cases correctly.
On the other hand, it might lead people into a trap, as their implementation is working fine wite dlt-daemon and may fail with others.

@LukeTheEngineer : I guess every code line which has a strlen() and "+1" might be suspicious:
./src/lib/dlt_user.c:1949:    const uint16_t name_size = (name != NULL) ? strlen(name)+1 : 0;
./src/lib/dlt_user.c:2055:            name_size = (varinfo->name != NULL) ? strlen(varinfo->name)+1 : 0;
./src/lib/dlt_user.c:2056:            unit_size = (varinfo->unit != NULL) ? strlen(varinfo->unit)+1 : 0;
./src/lib/dlt_user.c:2619:    const uint16_t name_size = (name != NULL) ? strlen(name)+1 : 0;
./src/lib/dlt_user.c:2637:    size_t str_truncate_message_length = strlen(STR_TRUNCATED_MESSAGE) + 1;

There was a fix in the past regarding ECU, APP, CTX IDs. They were not correctly handled by the test code in the past. So same mistake is also happening to dlt code itself.

> I've also removed +1 from 2621.
I do not see that patch yet. Have you forgot to push it? Please also do same rework for the TRUNCATED_MESSAGE and name_size.

After that, a lot of unit tests will fail. So those needs to be reworked as well. I am still not sure if it is all worth the effort.

LukeTheEngineer: @michael-methner since no major issues have been reported on the repo, I don't think the NULL termination presence will be as critical.

However, I will begin working on some changes and ensure the specification matches the product. 

To reiterate, I'm still new to the dlt daemon, and I am trying to understand how everything works when it comes to testing and doing general building. 

Thanks.

LukeTheEngineer: @michael-methner Could you tell me what tests you ran?

michael-methner: Hello @LukeTheEngineer ,
i checked the unit tests with: 
cmake .. -DCMAKE_BUILD_TYPE=Debug -DWITH_DLT_UNIT_TESTS=ON 
make -j16                                                                                                                                                                                                                               ctest *

Same thing is done by GitHub CI. You can see the results in "Check" tab of this PR.

LukeTheEngineer: @michael-methner , @mbehr1 Yep, the unit tests failed for me too. Specifically, the `gtest_dlt_user` caused a segfault due to the lack of null terminators, and `gtest_dlt_daemon_gateway` for unknown reasons.

I do have an idea for a fix, so if you would like to hear my in depth explanation, please do not hesitate to reach out.

michael-methner: Hello @LukeTheEngineer ,
i do not have a custom dlt_gateway.conf at my hand. But with default one, the test should pass. At least that is the state on the master branch:
```
5: [==========] 46 tests from 20 test suites ran. (3 ms total)                                                                                                             5: [  PASSED  ] 46 tests.                                                                                                                                           
5/8 Test #5: gtest_dlt_daemon_gateway ..................   Passed    1.63 sec
```

minminlittleshrimp: Hello @lvklevankhanh 
Please kindly have a look.

Hello @mbehr1 and @LukeTheEngineer ,
thanks for your interest in dlt topic.
To me, I agree with Michael's perspective. R22-11 quite new and we all need time to adapt.
You could easily see that huge things change after R21-11, especially the introduction of v2 protocol forcing dlt team to think about a Major release where we need to reimplement base header with HTYP2. We even do not prepare any plan for this change yet, so it will take time.
We might let this PR for a little bit later when we plan for the Major release.
Thanks for your understanding.
Minh

mbehr1: @minminlittleshrimp  I don't agree. This change/bugfix should be done completely independent of any htyp2 support. It's a smallish bugfix.

minminlittleshrimp: Hello @mbehr1 
Yes it is a small change, but it will affect our whole test suites which were implemented based on the old standard. Hence it requires a lot of effort to fix the whole test suites, which cannot be done in a few weeks. Plus we even do not know if there will be any further issues occur if this change is merged. Unit tests could not tell much things but some regression tests and smoketests might fail pointing to whatever legacy code inside dlt. Honestly, I think there must be some reasons that previous dlt maintainers decided to use Null like that. We will try our best to figure out and find the best way soon.
Thank you

mbehr1: @minminlittleshrimp I think changing the tests would be easy as well. But I just checked the old spec. So Autosar R4.0 Rev3 (spec v1.2.0) specifies:
Length of string + Unsigned 16-bit integer termination char.
Data string: Null terminated data string

So it's really a recent spec change. I wonder why this spec change was even approved. How is the review process for that?

Changing existing requirements so significantly is a mess...

So I'm fine in putting this in a new release where e.g. htyp2 will be supported as well. The trouble for v1 header will anyhow still be the same... (so tools need to handle zero-term and non-zero-term strings...)

michael-methner:
@mbehr1 , @LukeTheEngineer , @minminlittleshrimp 
Thank you for working on this PR. I close this PR now as we do not implement DLTv2 protocol as of today. DLT with respect to the NULL-terminator, we are in line with the old spec R20-11.

Issue #523, opened 2023-08-19
test failure on big endian platforms


LocutusOfBorg:
Hello, looks like we are failing testsuite on s390x... Is anything needed to debug the issue?
https://buildd.debian.org/status/fetch.php?pkg=dlt-daemon&arch=s390x&ver=2.18.10-1&stamp=1692269034&raw=0

minminlittleshrimp:
Hello @LocutusOfBorg 
Could you kindly share with me the dev environment information? I mean the debian version you are performing the test.
And do I need to setup anything else for system or just build dlt and trigger dlt unit tests?
I will try to reproduce the issue on docker.
Thank you

LocutusOfBorg:
Hello @minminlittleshrimp , this is my rules file (with cmake configuration)
https://sources.debian.org/src/dlt-daemon/2.18.10-3/debian/rules/
	dh_auto_configure -- \
		-DWITH_SYSTEMD=ON -DWITH_SYSTEMD_JOURNAL=ON -DWITH_MAN=ON -DWITH_DLT_ADAPTOR=ON \
		-DWITH_DLT_DBUS=ON -DWITH_DLT_SYSTEM=ON -DWITH_DLT_KPI=ON -DWITH_DLT_FILETRANSFER=ON -DWITH_UDP_CONNECTION=ON \
		-DWITH_DLT_UNIT_TESTS=ON -DWITH_EXTENDED_FILTERING=ON -DWITH_DLT_LOGSTORAGE_GZIP=ON -DCMAKE_LIBRARY_PATH=$(DEB_HOST_MULTIARCH)

minminlittleshrimp:
Hello @LocutusOfBorg 
Thanks for the info.
I could not access to the link, it keeps loading, so could you kindly share a raw txt file?
Regards

LocutusOfBorg:
```
#!/usr/bin/make -f
# See debhelper(7) (uncomment to enable)
# output every command that modifies files on the build system.
export DH_VERBOSE = 1

# see FEATURE AREAS in dpkg-buildflags(1)
#export DEB_BUILD_MAINT_OPTIONS = hardening=+all

# see ENVIRONMENT in dpkg-buildflags(1)
# package maintainers to append CFLAGS
#export DEB_CFLAGS_MAINT_APPEND  = -Wall -pedantic
# package maintainers to append LDFLAGS
#export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed


%:
[...]

PR #526, opened 2023-08-22
unittest: Fail with big-edian architecture 

LocutusOfBorg:
Swap some bytes makes it work on ppc64 and s390x

See: https://github.com/COVESA/dlt-daemon/issues/523

### DIFF
--- a/tests/gtest_dlt_daemon_gateway.cpp
+++ b/tests/gtest_dlt_daemon_gateway.cpp
@@ -458,6 +458,10 @@ TEST(t_dlt_gateway_parse_get_log_info, normal)
     uint8_t status = 7;
     uint16_t count_app_ids = 1;
     uint16_t count_context_ids = 1;
+#if (BYTE_ORDER == BIG_ENDIAN)
+    count_app_ids = DLT_HTOLE_16(count_app_ids);
+    count_context_ids = DLT_HTOLE_16(count_context_ids);
+#endif
     const char *apid = "LOG";
     const char *ctid = "TEST";
     const char *com = "remo";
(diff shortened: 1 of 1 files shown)


minminlittleshrimp:
Okay let me check the change with s390x/debian on docker
Regards

minminlittleshrimp:
Hello @LocutusOfBorg ,
thanks for your interest in DLT topic.
I could reproduce successfully in my local qemu.
Honestly, I think this FAIL of the test case is understandable, and as you pointed out, yes, it is BYTE_ORDER issue when the author of the test case was not aware of the Big edian case.
I will upload my fix together with an CI/CD for Big Edian check in the future.
Once again thank you so much.
Minh

tests/gtest_dlt_daemon_gateway.cpp
```diff
@@ -458,6 +458,10 @@ TEST(t_dlt_gateway_parse_get_log_info, normal)
     uint8_t status = 7;
     uint16_t count_app_ids = 1;
     uint16_t count_context_ids = 1;
+#if (BYTE_ORDER == BIG_ENDIAN)
+    count_app_ids = DLT_HTOLE_16(count_app_ids);
+    count_context_ids = DLT_HTOLE_16(count_context_ids);
+#endif
```
michael-methner, line 464: Hello @LocutusOfBorg ,
thanks for the patch but these IDs should not be in little endian when the MSBF bit is set.

I think we need rather a fix in the function dlt_getloginfo_conv_ascii_to_uint16_t() to reflect the endianess of the payload (as indicated by the MSBF bit)

minminlittleshrimp:
Close and move discussion to https://github.com/COVESA/dlt-daemon/pull/529

PR #529, opened 2023-08-23
unittest: apply big-edian case for converting ascii to uint16_t 

minminlittleshrimp:
When converting ascii to unint16_t, the byte order needs to be kept for big-edian case
Reworked from: https://github.com/COVESA/dlt-daemon/pull/526

### DIFF
--- a/tests/gtest_dlt_daemon_gateway.cpp
+++ b/tests/gtest_dlt_daemon_gateway.cpp
@@ -456,8 +456,8 @@ TEST(t_dlt_gateway_parse_get_log_info, normal)
     char ecuid[] = "ECU2";
     uint32_t sid = DLT_SERVICE_ID_GET_LOG_INFO;
     uint8_t status = 7;
-    uint16_t count_app_ids = 1;
-    uint16_t count_context_ids = 1;
+    uint16_t count_app_ids = DLT_HTOLE_16(1);
+    uint16_t count_context_ids =DLT_HTOLE_16(1);
     const char *apid = "LOG";
     const char *ctid = "TEST";
     const char *com = "remo";
@@ -531,6 +531,9 @@ TEST(t_dlt_gateway_parse_get_log_info, normal)
 
     msg.standardheader = (DltStandardHeader *)(msg.headerbuffer + sizeof(DltStorageHeader));
     msg.standardheader->htyp = DLT_HTYP_WEID | DLT_HTYP_WTMS | DLT_HTYP_UEH | DLT_HTYP_PROTOCOL_VERSION1;
+#if (BYTE_ORDER == BIG_ENDIAN)
+    msg.standardheader->htyp = (msg.standardheader->htyp | DLT_HTYP_MSBF);
+#endif
     msg.standardheader->mcnt = 0;
 
     dlt_set_id(msg.headerextra.ecu, ecuid);


minminlittleshrimp:
Tested with docker qemu: multiarch/ubuntu-core:s390x-focal
```
root@lum3hc-VirtualBox:/dlt-daemon/build# ctest
Test project /dlt-daemon/build
    Start 1: gtest_dlt_common
1/8 Test #1: gtest_dlt_common ..........................   Passed    0.19 sec
    Start 2: gtest_dlt_user
2/8 Test #2: gtest_dlt_user ............................   Passed    6.10 sec
    Start 3: gtest_dlt_daemon_common
3/8 Test #3: gtest_dlt_daemon_common ...................   Passed    0.05 sec
    Start 4: dlt_env_ll_unit_test
4/8 Test #4: dlt_env_ll_unit_test ......................   Passed    0.03 sec
    Start 5: gtest_dlt_daemon_gateway
5/8 Test #5: gtest_dlt_daemon_gateway ..................   Passed    1.86 sec
    Start 6: gtest_dlt_daemon_offline_log
6/8 Test #6: gtest_dlt_daemon_offline_log ..............   Passed    0.64 sec
    Start 7: gtest_dlt_daemon_event_handler
7/8 Test #7: gtest_dlt_daemon_event_handler ............   Passed    1.05 sec
    Start 8: gtest_dlt_daemon_multiple_files_logging
8/8 Test #8: gtest_dlt_daemon_multiple_files_logging ...   Passed    0.04 sec

100% tests passed, 0 tests failed out of 8

Total Test time (real) =  10.07 sec
root@lum3hc-VirtualBox:/dlt-daemon/build# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.3 LTS
Release:	20.04
Codename:	focal
root@lum3hc-VirtualBox:/dlt-daemon/build# lscpu | grep Byte*
Byte Order:                      Big Endian
root@lum3hc-VirtualBox:/dlt-daemon/build# 
```

LocutusOfBorg:
so the test actually caught a real bug?

minminlittleshrimp:
Yes, current implementation is unaware of big edian and assumes that logs comming from passive nodes will all be in little edian.

LocutusOfBorg:
amd64 is fine:
```
100% tests passed, 0 tests failed out of 9
```
s390x is fine:
```
100% tests passed, 0 tests failed out of 8
```
```

**Comment by LocutusOfBorg** [comment]
https://buildd.debian.org/status/fetch.php?pkg=dlt-daemon&arch=s390x&ver=2.18.10-5&stamp=1692816584&raw=0
it failed...

**Comment by LocutusOfBorg** [comment]
but maybe it's a flaky unrelated test failure

**Comment by LocutusOfBorg** [comment]
yeah, it works now!

**Comment by minminlittleshrimp** [comment]
Yup, I try to reproduce on my qemu but not succeed, the tests all pass.

File: src/shared/dlt_common.c [hunk 1]
```diff
@@ -4149,13 +4149,20 @@ int16_t dlt_getloginfo_conv_ascii_to_uint16_t(char *rp, int *rp_count)
     if ((rp == NULL) || (rp_count == NULL))
         return -1;
 
+#if (BYTE_ORDER == BIG_ENDIAN)
+    num_work[0] = *(rp + *rp_count + 0);
+    num_work[1] = *(rp + *rp_count + 1);
+    num_work[2] = *(rp + *rp_count + 2);
+    num_work[3] = *(rp + *rp_count + 3);
+#else
```
michael-methner, line 4157: The endianess of the payload of the getinfolog message is not about the host endianness (as indicated by the macro BYTE_ORDER). The MSBF bit of the dlt message have to be evaluated here. Unfortunately it is not pased down to this function.

As per autosar spec R20-11 Log and Trace Protocol Specification:
[PRS_Dlt_00604] If the MSBF bit is set to ‘0’, the payload data is encoded in little endian format. ⌋ (RS_LT_00002, RS_LT_00016)
[PRS_Dlt_00605] If the MSBF bit is set to ‘1’, the payload data is encoded in big endian format. ⌋ (RS_LT_00002, RS_LT_00016)

Unfortunately, the information of the MSBF bit is not passed to this function. As the function is part of dlt_common.h and therefore part of the public DLT-API, we cannot change it without major version update of DLT. I would propose to add new functions like:
dlt_getloginfo_conv_ascci_to_*_msbf(char *rp, int *rp_count, int msbf);
to maintain backward API compatiblity.

@minminlittleshrimp , please give me a call before you start implementing.

michael-methner:
Hello @LocutusOfBorg , hello @minminlittleshrimp ,
this is the third PR for fixing the same bug. Shall I close #526 and we continue discussion here?

michael-methner:
> so the test actually caught a real bug?

@LocutusOfBorg : Yes, but to be honest, after seeing the implementation here, I doubt that dlt-daemon will correctly run on big-endian machines even when the unit tests are passing. Getting reliable support for big endian machines will require a lot of testing and potentially a lot of bug fixing. 

Anyhow, I do not know any recent Linux embedded platform using big-endian. So it is not a major deal for me.

src/shared/dlt_common.c
michael-methner: Hello @minminlittleshrimp , hello @LocutusOfBorg ,

i would like to be a bit more pragmatic here. Lets keep the payload as little endian for big and little endian machines.
Then we could convert all 16bit- values in the unit test using htole16() and we are done and unblock the build of the  Debian packages.

Adding now an API-change would delay the solution until we release a DLT 3.0 (which is not even on the roadmap).

I would keep a bug ticket open that big-endian payloads are not supported.

What do you think?

LocutusOfBorg:
@michael-methner as example dlt-daemon is used by mosquitto, and mosquitto runs everywhere, including some BE platforms (well, just the library is used, for sure not the gateway!)

src/shared/dlt_common.c
minminlittleshrimp: Hello @michael-methner 
I think it is good enough for now.
I soon apply the solution for review.
Regards

michael-methner (approved):

PR #530, opened 2023-08-24
[DRAFT] src: refactoring current implementation

minminlittleshrimp:
+ Remove dead code
+ Add NULL-variable and failed-process detection
+ Assign delimiters to const char type
+ Handle multithread cancellation

### DIFF
--- a/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
+++ b/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
@@ -622,7 +622,10 @@ int dlt_logstorage_open_log_file(DltLogStorageFilterConfig *config,
                          "%s: Remove '%s' (num_log_files: %u, config->num_files:%u)\n",
                          __func__, absolute_file_path, num_log_files,
                          config->num_files);
-                remove(absolute_file_path);
+
+                if (remove(absolute_file_path) != 0)
+                    dlt_log(LOG_ERR, "Could not remove file\n");
+
                 free((*head)->name);
                 (*head)->name = NULL;
                 *head = n->next;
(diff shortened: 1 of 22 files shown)


src/offlinelogstorage/dlt_offline_logstorage_behavior.c
```diff
@@ -622,7 +622,10 @@ int dlt_logstorage_open_log_file(DltLogStorageFilterConfig *config,
                          "%s: Remove '%s' (num_log_files: %u, config->num_files:%u)\n",
                          __func__, absolute_file_path, num_log_files,
                          config->num_files);
-                remove(absolute_file_path);
+
+                if (remove(absolute_file_path) != 0)
```
lvklevankhanh, line 626: Please apply this fix to check return at line #584

lvklevankhanh:
Please check the line https://github.com/COVESA/dlt-daemon/blob/master/src/shared/dlt_multiple_files.c#L103
Should it be "newest" instead?

lvklevankhanh:
Should we check file null before using at line?
https://github.com/COVESA/dlt-daemon/blob/master/src/offlinelogstorage/dlt_offline_logstorage_behavior.c#L430
For reference: 
https://github.com/COVESA/dlt-daemon/blob/22d394f4a633c7f08e13fd926339b879cb6f64da/src/kpi/dlt-kpi-common.c#L61

lvklevankhanh:
Should we fix the coverity "Calling "malloc" which returns uninitialized memory" at lines?
https://github.com/COVESA/dlt-daemon/blob/master/src/daemon/dlt_daemon_common.c#L934
https://github.com/COVESA/dlt-daemon/blob/master/src/daemon/dlt_daemon_common.c#L946

minminlittleshrimp:
Set to draft since lower prior

minminlittleshrimp:
Close PR and replace by rebased PR: https://github.com/COVESA/dlt-daemon/pull/566

PR #531, opened 2023-08-29
fix: Contact delimiter to log file name

zhangzhiyu19850809:
There may be garbled codes after the delimiter in the offline log file name when concatenate the delimiter to the offline log file name if the delimiter' next char is not '\0' in memory.


michael-methner (approved):
Looks good to me. thanks for the patch

michael-methner:
Hello @zhangzhiyu19850809 ,
this patch conflicts with a larger rework on logstorage which was just applied to the master branch.
Anyhow your issue should be solved with that implementation as it is now using sprintf + "%c".

Can you please check? Then I will close this PR

zhangzhiyu19850809:
> Hello @zhangzhiyu19850809 , this patch conflicts with a larger rework on logstorage which was just applied to the master branch. Anyhow your issue should be solved with that implementation as it is now using sprintf + "%c".
> 
> Can you please check? Then I will close this PR

Hi,@michael-methner],thanks for your reply. Since the issue is solved on the master branch, you can surely close the PR.

PR #525, opened 2023-08-22
Logstorage changes

lvklevankhanh:

PR #542, opened 2023-09-26
Revert "Logstorage changes"

minminlittleshrimp:
Reverts COVESA/dlt-daemon#525

I would like to revert and take charge of this logstorage porting.
It becomes so complicated to rework on the latest commit where everything of logstorage unit tests are not well-ported and the code base gets worse.


lvklevankhanh (approved):

minminlittleshrimp:
Should not revert but just rework due to many commits pushed. Hence close and stick to rework minimal parts of this feature

PR #544, opened 2023-09-28
Add common dlt_cdh_cpuinfo.c to unblock build on non amd64 and i386 arcs

LocutusOfBorg:


michael-methner:
Hello @LocutusOfBorg ,
thanks for the patch but what are you trying to achieve here? I think it does not make sense to make this just compilable on other platforms.  

The coredump handler was never used by us (ADIT) and I am wondering if there is still interest in maintaining this part by someone in times of minicoredumper. 

@DLT community, @lhelwing, @oncaphillis, @stefanvacek-intel : Is is the DLT coredump handler still used by anyone?  Otherwise I would tend to remove it completely from DLT.

michael-methner (approved):

michael-methner:
Hello @LocutusOfBorg ,
thank you for clarification. I go on with merging as you are using this subcomponent actively.  👍

LocutusOfBorg:
thanks!

PR #547, opened 2023-10-05
Add logstorage exclusion feature

ranoongs:
This commit is to add 'exclusion' feature in logstorage. It can be used when taking in all the DLT logs and exclude some modules (Using wildcards to App ID and Context ID in the filter).


minminlittleshrimp:
Hello @ranoongs 
FYI, the upstream is available for review now at: https://github.com/COVESA/dlt-daemon/pull/557

PR #553, opened 2023-10-25
Add Journal Gateway Mode

harisokanovic:
Add Journal Gateway Mode

Adds a new connection type which reads the Systemd Journal log stream
for current boot and forwards everything to connected clients. This new
mode is useful on Linux system that predominantly use the Journal for
logging, which dlt-daemon cannot replace outright. It's enabled by
adding 'JournalGatewayMode = 1' to dlt.conf.

This mode of dlt-daemon is similar to dlt-system's Journal Adapter,
enabled by 'JournalEnable = 1' in dlt-system.conf. However, this new
implementation preserves certain Journal meta-data in DLT formatting
to facilitate filtering via control messages:

* App ID is the first 4 alpha-numeric chars of SYSLOG_IDENTIFIER.
  Examples:
    "httpd" ==> "HTTP"
    "wpa_supplicant" ==> "WPAS"
    "ls" ==> "LS<NULL><NULL>"

* Context ID is translated from SYSLOG_FACILITY enum.
  Examples:
    LOG_KERN ==> "KERN"
    LOG_DAEMON ==> "DAEM"
    LOG_FTP ==> "FTP<NULL>"

* Level is translated from PRIORITY enum as follows:
    LOG_EMERG, LOG_ALERT, LOG_CRIT ==> DLT_LOG_FATAL
    LOG_ERR ==> DLT_LOG_ERROR
    LOG_WARNING ==> DLT_LOG_WARN
    LOG_NOTICE, LOG_INFO ==> DLT_LOG_INFO
    LOG_DEBUG ==> DLT_LOG_DEBUG

Level filtering in this mode is implemented by the Journal read loop.
Each log entry is checked against a hash map of App ID to limit or the
default if not present in map. The hash map and default can be updated
by control messages from clients.


minminlittleshrimp:
Instead of adding this new mode, is there anyway we could improve the current implementation of dlt-system?
I do not think adding new mode for just only filtering purpose is a good idea.

harisokanovic:
> Instead of adding this new mode, is there anyway we could improve the current implementation of dlt-system?
> I do not think adding new mode for just only filtering purpose is a good idea.

Hi Minh,

I went with the current approach because I think it performs better than any modification of dlt-system:

If I understand the architecture correctly, dlt-system would need to create multiple app/context IDs to dlt-daemon (e.g. via libdlt.so) to give every system process a unique ID in the DLT protocol. This incurs unnecessary bookkeeping overhead.

Further complicating booking is that Journal log entries are stateless. It's difficult to know when a process exists (vs when it's quiet) to tear down the aforementioned contexts.

The proposed approach in dlt-daemon is stateless. The daemon doesn't have to consult nor update any per-process bookkeeping data to process a log entry. It simply translates meta-data from Journal fields to DLT fields - i.e. a "gateway" between the two logging systems.

The one caveat is filtering: However filtering overhead is proportional to the number filter rules configured by clients, and not the total number of processes running on the system. This benefits performance because the filter set is smaller (usually 0) than the number of running processes.

I hope the above explanation make sense. Please let me know if I'm misunderstanding the architecture in some way.

michael-methner:
Hello @harisokanovic ,
besides the fact that your implementation is well done, i think i have to second @minminlittleshrimp  arguments and would prefer a change to the existing mechanism in dlt-system for following reasons:

1) dlt-daemon runs on QNX and Android as well which do not have sdjournal. Also on those platforms we rely on external programs interfacing with the native logging mechansims (logd for Android and slog for QNX)
2) The concept of stateless logging does not fit to the DLT protocol. DLT relies here on registered app ids and context ids which are also maintained in the DltDaemonRegisteredUsers structure. I do not know if your implementation would break any existing code related to this.
3) the automatic translation to app ids might cause conflicts with already registered application which might lead to ambiguities. 

Please let me know your thoughts.

michael-methner:
No response from author since 3 month -> closing this PR

PR #554, opened 2023-10-25
Add ECUSoftwareVersionFileField option to dlt.conf

harisokanovic:
Add ECUSoftwareVersionFileField option to dlt.conf

Modify dlt_daemon_local_ecu_version_init() to read a single field from
KEY=VALUE (or quoted KEY="VALUE") style files, so that dlt-daemon can
report an OS version number parsed from /etc/os-release file.

For example, get_software_version would return "22.04.3 LTS" under the
following configuration:

$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.3 LTS"
NAME="Ubuntu"
VERSION="22.04.3 LTS"
VERSION_CODENAME=jammy

$ cat /etc/dlt.conf
PathToECUSoftwareVersion = /etc/os-release
ECUSoftwareVersionFileField = VERSION

### DIFF
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -389,6 +389,7 @@ int option_file_parser(DltDaemonLocal *daemon_local)
     daemon_local->daemonFifoSize = 0;
     daemon_local->flags.sendECUSoftwareVersion = 0;
     memset(daemon_local->flags.pathToECUSoftwareVersion, 0, sizeof(daemon_local->flags.pathToECUSoftwareVersion));
+    memset(daemon_local->flags.ecuSoftwareVersionFileField, 0, sizeof(daemon_local->flags.ecuSoftwareVersionFileField));
     daemon_local->flags.sendTimezone = 0;
     daemon_local->flags.offlineLogstorageMaxDevices = 0;
     daemon_local->flags.offlineLogstorageDirPath[0] = 0;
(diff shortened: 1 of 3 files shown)


src/daemon/dlt-daemon.c
```diff
@@ -1906,78 +1914,143 @@ int dlt_daemon_local_connection_init(DltDaemon *daemon,
     return 0;
 }
 
-int dlt_daemon_local_ecu_version_init(DltDaemon *daemon, DltDaemonLocal *daemon_local, int verbose)
+static char* file_read_everything(FILE* const file, const size_t sizeLimit)
 {
-    char *version = NULL;
-    FILE *f = NULL;
-
-    PRINT_FUNCTION_VERBOSE(verbose);
```
minminlittleshrimp, line 1914: Why remove this macro? It is required for internal dlt logs when Verbose = 1

harisokanovic: I didn't remove it. Line 1914 got moved to line 2031 (below):

```
+ int dlt_daemon_local_ecu_version_init(DltDaemon *daemon, DltDaemonLocal *daemon_local, int verbose)
+ {
+    FILE *f = NULL;
+
+    PRINT_FUNCTION_VERBOSE(verbose);
```

michael-methner (approved):

PR #557, opened 2023-10-30
Structure for negative filter logstorage

minminlittleshrimp:
Upstream negative filter proprietary


lvklevankhanh (approved):

minminlittleshrimp:
See: https://github.com/COVESA/dlt-daemon/pull/547
This is the upstream proprietary from DLT.

PR #561, opened 2023-11-06
dlt-receive: add a -r option to automatically reconnect the socket wh…

LocutusOfBorg:
…en it drops

### DIFF
--- a/src/console/dlt-receive.c
+++ b/src/console/dlt-receive.c
@@ -119,6 +119,7 @@ typedef struct {
     int vflag;
     int yflag;
     int uflag;
+    int rflag;
     char *ovalue;
     char *ovaluebase; /* ovalue without ".dlt" */
     char *fvalue;       /* filename for space separated filter file (<AppID> <ContextID>) */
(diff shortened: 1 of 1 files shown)


lvklevankhanh:
Hello @LocutusOfBorg 
I tried your changes and it doesn't work from my side with the command "dlt-receive -r -a localhost". It works with the sample command "dlt-receive -a localhost". Please help me verify from your side.
Regards,

minminlittleshrimp:
Hello @LocutusOfBorg 
Could you kindly explain/provide a specific scenario for us?
We could not clearly understand this PR.
Why we have to loop the piece of code used for verbose mode logging of dlt-receive?
Regards

LocutusOfBorg:
Hello, this happens when I try to connect to an old version of dlt-daemon (forked from 2.9.0), with the same dlt-receive tool. Probably something in the libdlt changed in the meanwhile, and that connect is now proof to connection drops.
We can close, I can use a newer dlt-receive and live happy (it's common to use the one built with the same yocto recipe, because it's already there, but not big deal) closing!

PR #567, opened 2023-11-13
dlt_cdh:

LocutusOfBorg:
Make sure on 64 bit we read an ELF64 structure.
Otherwise we get a read error, and the context file is missing some good to know information


michael-methner:
Hello @LocutusOfBorg ,
thanks for the contribution.

PR #574, opened 2023-11-23
systemd: Don't install services in basic.target, but use multi-user.t…

LocutusOfBorg:
…arget instead.

The specified systemd service file declares an unusual WantedBy= relationship.

Most services that want to be started automatically at boot should use WantedBy=multi-user.target or WantedBy=graphical.target. Services that want to be started in rescue or single-user mode should instead use WantedBy=sysinit.target

Please refer to https://wiki.debian.org/Teams/pkg-systemd/rcSMigration for details.

### DIFF
--- a/systemd/dlt-dbus.service.cmake
+++ b/systemd/dlt-dbus.service.cmake
@@ -27,5 +27,5 @@ NotifyAccess=main
 LimitCORE=infinity
 
 [Install]
-WantedBy=basic.target
+WantedBy=multi-user.target
 
--- a/systemd/dlt-system.service.cmake
+++ b/systemd/dlt-system.service.cmake
@@ -27,5 +27,5 @@ NotifyAccess=main
 LimitCORE=infinity
 
 [Install]
-WantedBy=basic.target
+WantedBy=multi-user.target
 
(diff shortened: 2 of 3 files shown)


minminlittleshrimp:
dlt should be booted in basic.target phase when everything is setup.
sysinit ->[timers, paths, sockets]->basic
Of course, /tmp must be mounted first for socket creation.
Suggestion of dlt.service:

...
DefaultDependencies=no ==> this means After=basic.target
After=tmp.mount
[Service]
...
~~[Install]~~
~~WantedBy=<whatever>~~

Moreover:
```
multi-user.target[¶](https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#multi-user.target)

    A special target unit for setting up a multi-user system (non-graphical). This is pulled in by graphical.target.

    Units that are needed for a multi-user system shall add Wants= dependencies for their unit to this unit during installation. This is best configured via WantedBy=multi-user.target in the unit's [Install] section.
```
We do not need any dependency here in dlt.service

Regards

LocutusOfBorg:
Can you please followup with your changes then?

minminlittleshrimp:
Since I cannot edit your repo, I close this one and continue with mine in which I add you as coauthor.
Regards

PR #575, opened 2023-11-27
systemd: correct service files relationship

minminlittleshrimp:
+ Remove unusual relationship and replace with multi-user
+ Avoid dlt conflict on shutdown when applying no dependency
See https://github.com/COVESA/dlt-daemon/pull/574

### DIFF
--- a/systemd/dlt.service.cmake
+++ b/systemd/dlt.service.cmake
@@ -16,6 +16,10 @@
 [Unit]
 Description=COVESA DLT logging daemon
 Documentation=man:dlt-daemon(1) man:dlt.conf(5)
+DefaultDependencies=no
+Before=shutdown.target
+Conflicts=shutdown.target
+After=tmp.mount
 
 [Service]
 Type=simple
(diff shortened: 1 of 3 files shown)


LocutusOfBorg:
thanks!

lvklevankhanh (approved):

systemd/dlt.service.cmake
```diff
@@ -24,6 +26,3 @@ ExecStart=@CMAKE_INSTALL_PREFIX@/bin/dlt-daemon
 WatchdogSec=@DLT_WatchdogSec@
 NotifyAccess=main
 LimitCORE=infinity
-
-[Install]
-WantedBy=basic.target
```
michael-methner, line 29: Hello @minminlittleshrimp ,
this disables the start of dlt. Did you want to shift this one also to multi-user.target?

minminlittleshrimp: Hello @michael-methner 
This should not be shifted since daemon should be up before dlt-system or any in multi-user.
I will correct it.

minminlittleshrimp:
Hello @michael-methner 
Kindly consider merging this PR for me.
Thank you

michael-methner (approved):

Issue #533, opened 2023-09-04
ringbuffer of dlt-daemon only record the old log messsages


cumtsmart:
In my opinion, the ringbuffer of dlt-daemon will record the latest dlt log if no dlt-client connect.  when the buffer is full, the new coming log will squeeze out the oldest one, this's what ringbuffer means. But i found ringbuffer just drop the new coming dlt log when buffer is full.

cumtsmart:
`    if ((sock != DLT_DAEMON_SEND_FORCE) &&
        ((daemon->state == DLT_DAEMON_STATE_BUFFER) || (daemon->state == DLT_DAEMON_STATE_SEND_BUFFER) ||
         (daemon->state == DLT_DAEMON_STATE_BUFFER_FULL))) {
        if (daemon->state != DLT_DAEMON_STATE_BUFFER_FULL) {
            /* Store message in history buffer */
            ret = dlt_buffer_push3(&(daemon->client_ringbuffer), data1, size1, data2, size2, 0, 0);
            if (ret < DLT_RETURN_OK) {
                dlt_daemon_change_state(daemon, DLT_DAEMON_STATE_BUFFER_FULL);
            }
        }
        if (daemon->state == DLT_DAEMON_STATE_BUFFER_FULL) {
            daemon->overflow_counter += 1;
            if (daemon->overflow_counter == 1)
                dlt_vlog(LOG_INFO, "%s: Buffer is full! Messages will be discarded.\n", __func__);

            return DLT_DAEMON_ERROR_BUFFER_FULL;
        }
    } else {
        if ((daemon->overflow_counter > 0) &&
            (daemon_local->client_connections > 0)) {
            sent_message_overflow_cnt++;
            if (sent_message_overflow_cnt >= 2) {
                sent_message_overflow_cnt--;
            }
            else {
                if (dlt_daemon_send_message_overflow(daemon, daemon_local,
                                          verbose) == DLT_DAEMON_ERROR_OK) {
                    dlt_vlog(LOG_WARNING,
                             "%s: %u messages discarded! Now able to send messages to the client.\n",
                             __func__,
                             daemon->overflow_counter);
                    daemon->overflow_counter = 0;
                    sent_message_overflow_cnt--;
                }
            }
        }
    }`

cumtsmart:
`        if (daemon->state != DLT_DAEMON_STATE_BUFFER_FULL) {
            /* Store message in history buffer */
[...]

PR #579, opened 2023-12-04
Delete the oldest entry when ringbuffer is full

jianjun-huang:
The change will delete the oldest entries in the ringbuffer when the ringbuffer is full


michael-methner:
Hello @jianjun-huang ,
there are reasons to not overwrite the first messages as well as there are reasons to have the ring buffer to be rotated.

Could you make this feature configurable via enviroment_variables (like the ringbuffer size already is)? The default behavior should be to keep the oldest logs. This enables to debug startup issues for example.

minminlittleshrimp:
Hello @jianjun-huang 
thanks for the idea.
However, as @michael-methner mentioned, we would not want our current mechanism changing, but refer a configurable feature of log rotating inside the ring buffer.
About the current mechanism, you could find more info in this disscussion: https://github.com/COVESA/dlt-daemon/issues/533
Please kindly rework this patch, we will review when it done.
Thank you

jianjun-huang:
Hi @minminlittleshrimp 
I'm trying to work on making the feature configurable. 

Just like the discussion in https://github.com/COVESA/dlt-daemon/issues/533, the same ringbuffer implementation is used in dlt-daemon and libdlt. It is easier to add a new configuration parameter in the dlt.conf to change the behavior on dlt-daemon site. But there is no good solution on the application side(libdlt). Environment variable is a potential application wide but not a system wide solution. The user may observe different ringbuffer behaviors among the applications and dlt-daemon.

Also, in the AUTOSAR_SWS_LogAndTrace_R22_11.pdf, it mentions:
"[SWS_LOG_00001] dLog message logged before the Logging framework is able
to process them (e.g. daemon communication is not established) shall be queued.
The queue size is defined by LogAndTraceInstantiation.queueSize. If this size
is exceeded the oldest entries shall be discarded.c(RS_LT_00003, RS_LT_00052)"

It seems that we should always remove the oldest entries when ringbuffer is full.

What are your thoughts on these questions?

minminlittleshrimp:
Hello @jianjun-huang 
Unlike LTTng, DLT tool in fact part of COMstack/Autosar, and hence, serving the purpose of logging and tracing mainly on Automotive side. DLT is designed and constructed under Autosar standard, so honestly if the doc forcing to discard new log, no ring buffer rotating, then this behavior will be and always be the default.
On my point of view, since dlt is OSS project, it can be enhanced the flexibility, reusability by adding new feature but still keeping the original/standard one. In this case, I agree with Michael to at least make the log rotating optional! Using env variable seems fine here, by setting a variable to make a rotation of log once at a time sounds good to me then.
Another suggestion is to look into the way logstorage buffer rotating with option overwrite behavior DISCARD_OLD.
Regards

minminlittleshrimp:
1 more point: "Environment variable is a potential application wide but not a system wide solution. The user may observe different ringbuffer behaviors among the applications and dlt-daemon."

I would say, libdlt ring buffer will be used by all users, and there already locking mechanism for the data safe, so I do not think that setting an environment variable will lead to different behavior. For instance,by export DLT_USER_RINGBUFFER_MAX=val
We increase the libdlt buffer to that value, and all users will use this buffer until the buffer full. So I think we can make the log rotating option using same way.

minminlittleshrimp:
Sorry @jianjun-huang 
We are quite busy and late response on this PR.
I am preparing my env at my local and testing this patch.
Will report my result later.

minminlittleshrimp:
This feature is cool and I dont want to drop it, but it here for so long and need rework.
--> let try on our side for the rework based on this work from @jianjun-huang

minminlittleshrimp:
Close feel free to reopen

PR #580, opened 2023-12-04
add functionality to dump ring buffer to disk file

jianjun-huang:
For example:
/usr/bin/dlt-logstorage-ctrl -D /tmp/ss.dlt

will dump the entries in ringbuffer to file /tmp/ss.dlt

### DIFF
--- a/src/console/logstorage/dlt-logstorage-common.h
+++ b/src/console/logstorage/dlt-logstorage-common.h
@@ -28,6 +28,7 @@
 #define EVENT_UNMOUNTING    0
 #define EVENT_MOUNTED       1
 #define EVENT_SYNC_CACHE    2
+#define EVENT_DUMP_RING_BUFFER    3
 
 typedef enum
 {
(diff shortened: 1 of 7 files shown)


minminlittleshrimp:
Hello @jianjun-huang 
thanks for your contribution, this looks interesting feature for dlt.
Dumping logs from ring buffer to files would be very useful in case no daemon available and we need to rotate the buffer for new logs. This prevents missing all the old logs.
I will review and test on my local.
Regards

minminlittleshrimp:
Hello @jianjun-huang 
I read through the code change and it seems the same functionality of our current implementation of Sync Behavior ON_DEMAND: When activate this sync option, we just need to send a sync signal for dumping logs from cache (logstorage ring buffer) to dlt files.
Could you kindly clarify my point?
In this case, we will reconsider to not approve for this PR.
Regards

jianjun-huang:
Hi @minminlittleshrimp 
this option  is similar to ON_DEMAND, but this option does not remove logs from ringbuffer, just peek dlt messages from the ringbuffer.

minminlittleshrimp:
I see, I need to check in my local. So sorry for the late response, we are busy these days releasing internal dlt.

minminlittleshrimp:
Close feel free to reopen

PR #583, opened 2023-12-08
Limit set of exported symbols

Felix-El:
This disables unconditional export of symbols from the shared library `libdlt` in favor of an opt-in approach.

- only symbols marked DLT_EXPORT are exposed from the shared library
- this commit adds DLT_EXPORT on all symbols declared in the installed header files
- static library builds of `libdlt` are unaffected by this change (`BUILD_SHARED_LIBS=OFF`)

```
# Build previous HEAD
$> nm --dynamic --defined-only ./build/src/lib/libdlt.so | cut -f3 -d' ' | sort > ./before
# Build this commit
$> nm --dynamic --defined-only ./build/src/lib/libdlt.so | cut -f3 -d' ' | sort > ./after
# Summary (Linux, with maximum options enable, -DDLT_IPC=UNIX_SOCKET)
$> wc -l ./before ./after
  331 ./before
  272 ./after
```

The sorted list of remaining symbols is:
```
$> cat ./after
dlt_buffer_check_size
... all prefixed with dlt ...
dlt_with_timestamp
get_filename_ext
getFileSerialNumber
multiple_files_buffer_file_name
... more multiple_files_buffer_* ...
multiple_files_buffer_write_chunk
```

So some functions need to be revised:

1.  `dlt_buffer_*` defined in `src/shared/dlt_common.c`
    These functions are required to be exported but they are not declared in any header.
2. `get_filename_ext`  declared in `include/dlt/dlt_common.h`
    It is not prefixed with *dlt*. Should it really be exported?
3. `getFileSerialNumber` defined in `src/lib/dlt_filetransfer.c`
    This function should be renamed to have a *dlt* prefix. It is also required to be exported but not declared in any header.
4. `multiple_files_buffer_*` declared in `include/dlt/dlt_multiple_files.h`
    These functions appear to be public API, too but are not prefixed with *dlt*.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@
 #######
 
 # Set minimum Cmake version and setup policy behavior
-cmake_minimum_required(VERSION 3.3)
+cmake_minimum_required(VERSION 3.12)
 
 if(${CMAKE_VERSION} VERSION_GREATER "3.20" OR ${CMAKE_VERSION} VERSION_EQUAL "3.20")
     cmake_policy(SET CMP0115 OLD)
--- a/include/dlt/CMakeLists.txt
+++ b/include/dlt/CMakeLists.txt
@@ -19,11 +19,15 @@ endif()
 
 configure_file(dlt_user.h.in dlt_user.h)
 
+include(GenerateExportHeader)
+generate_export_header(dlt)
+
 set(HEADER_LIST dlt.h dlt_user_macros.h dlt_client.h dlt_protocol.h
                 dlt_common.h dlt_types.h dlt_shm.h dlt_offline_trace.h
                 dlt_filetransfer.h dlt_common_api.h dlt_multiple_files.h
                 ${CMAKE_CURRENT_BINARY_DIR}/dlt_version.h
-                ${CMAKE_CURRENT_BINARY_DIR}/dlt_user.h)
+                ${CMAKE_CURRENT_BINARY_DIR}/dlt_user.h
+                ${CMAKE_CURRENT_BINARY_DIR}/dlt_export.h)
 
 if(WITH_DLT_DISABLE_MACRO)
     list(REMOVE_ITEM HEADER_LIST dlt_user_macros.h)
--- a/include/dlt/dlt_client.h
+++ b/include/dlt/dlt_client.h
@@ -73,6 +73,7 @@
  \{
  */
 
+#   include "dlt_export.h"
 #   include "dlt_types.h"
 #   include "dlt_common.h"
 #include <stdbool.h>
@@ -106,8 +107,8 @@ typedef struct
 extern "C" {
 #   endif
 
-void dlt_client_register_message_callback(int (*registerd_callback)(DltMessage *message, void *data));
-void dlt_client_register_fetch_next_message_callback(bool (*registerd_callback)(void *data));
+DLT_EXPORT void dlt_client_register_message_callback(int (*registerd_callback)(DltMessage *message, void *data));
+DLT_EXPORT void dlt_client_register_fetch_next_message_callback(bool (*registerd_callback)(void *data));
 
 /**
  * Initialising dlt client structure with a specific port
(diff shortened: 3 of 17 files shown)


alexmohr:
At least 4. should not be used via public API imho as these functions are used internally to have rolling file names. There is no point in exposing that via the public API as clients should not write files with the dlt library.
getFileSerialNumber is really necessary for public api? Shouldn't the public api just offer functions to transmit files w/o the user caring about how in detail the file is sent?

Felix-El:
> At least 4. should not be used via public API imho as these functions are used internally to have rolling file names. There is no point in exposing that via the public API as clients should not write files with the dlt library. getFileSerialNumber is really necessary for public api? Shouldn't the public api just offer functions to transmit files w/o the user caring about how in detail the file is sent?

The symbols I listed are not only public (user-facing) API. However, they have to be exported at the moment because some console tools expect them to be there. This is because `libdlt` is currently also used for sharing some code (implementation details) between daemon and other tools in the project.

Generally I'd propose to extract these details into a separate static library with private headers. To continue with this, however, I'd need some basic commitment to this PR's general idea by any one of the project responsible.

michael-methner:
Hello @Felix-El ,
thanks for your contribution. I think this really make sense. 

As this anyhow breaks ABI-compatiblity, I would also like to have the functionname #2 to #4 renamed.
Removing the common code for dlt-daemon and dlt-tools I would keep for a separate PR.

There is a version bump in cmake. So DLT will not be compilable on older platforms (e.g. Ubuntu 18.04). Can you make this changes backward compatible? E.g. by using this feature only on recent cmake versions?

Felix-El:
> As this anyhow breaks ABI-compatiblity, I would also like to have the functionname #2 to #4 renamed.
Yus, totally agree.

> Removing the common code for dlt-daemon and dlt-tools I would keep for a separate PR.
Fine for me.

> There is a version bump in cmake. So DLT will not be compilable on older platforms (e.g. Ubuntu 18.04). Can you make this changes backward compatible? E.g. by using this feature only on recent cmake versions?

Is it maybe an option to just copy https://github.com/Kitware/CMake/blob/v3.12.0/Modules/GenerateExportHeader.cmake into this repo? License is BSD 3-Clause so should not be an issue. I would just need to verify this generates fine even with CMake as old as 3.3.

Felix-El:
Hello everyone, I have just pushed an updated version which addresses above discussion points.

- `get_filename_ext` renamed to `dlt_get_filename_ext` to namespace it for the time being (may later be remove from export)
- `getFileSerialNumber` renamed to `dlt_get_file_serial_number` for the same reason
- `multiple_files_*` were not renamed but excluded from export
- tested to build with CMake down to v3.5.1 on Ubuntu 16.04 (GenerateExportHeader copied from v3.12.0)

Since there are many more API to revise for "should they be public?" I deemed it enough to rename the few non-dlt* prefixed exported functions so we have everything under a dlt* "namespace" for the time being. This already reduces the risk of conflict extremely and further improvement can be done after merging this.

minminlittleshrimp:
@Bichdao021195 
Please kindly check this rework of dlt symbol export

BichDaoBosch:
Hi @minminlittleshrimp 
I tested to build with CMake v3.3 on Ubuntu 20. 
```
 Build previous HEAD
$> nm --dynamic --defined-only ./build/src/lib/libdlt.so | cut -f3 -d' ' | sort > ./before
# Build this commit
$> nm --dynamic --defined-only ./build/src/lib/libdlt.so | cut -f3 -d' ' | sort > ./after

$> wc -l ./before ./after
314 ./before
  258 ./after
  572 total
```
With cat ./after , i can observed that 
get_filename_ext renamed to dlt_get_filename_ext ->ok
getFileSerialNumber renamed to dlt_get_file_serial_number -> ok
multiple_files_* were not renamed but excluded from export -> ok
tested to build with CMake  v3.3 on Ubuntu 20  -> able to be exported 
Thanks 
Dao Nguyen

Felix-El:
Does this mean this can be closed/merged? Any commit that is merged prior to this might cause another rebase and review iteration (i.e. if new functions are added in the meantime).

PR #593, opened 2024-01-31
Remove extra file number increment in name

### LINKED ISSUE #592: DLT Receiver rename error and possible file loss
I have a device that transmits DLT traces and want to capture them. This device sometimes disconnects DLT so I have to reconnect. I am also using a file size limit. When DLT disconnects and the receiver reconnects to it, after reaching the size limit, the rename of the file has an additional increment to the number. I have built this project with cygwin so the command I use is the following:

dlt-receive.exe -a -o .\log.dlt -c 200M -p 3490 192.168.0.2

Steps to reproduce:
1. Run from cmd with a suitable size limit: dlt-receive.exe -a -o .\log.dlt -c 200M -p 3490 192.168.0.2
2. Make the device restart so the connection drops
3. When restarted, run again: dlt-receive.exe -a -o .\log.dlt -c 200M -p 3490 192.168.0.2
4. Wait until size limit is reached

This should be reproducable in a linux environment as well

Laximas:
Fix #592

### DIFF
--- a/src/console/dlt-receive.c
+++ b/src/console/dlt-receive.c
@@ -284,14 +284,17 @@ int dlt_receive_open_output_file(DltReceiveData *dltdata)
         char filename[PATH_MAX + 1];
         filename[PATH_MAX] = 0;
 
-        snprintf(filename, PATH_MAX, "%s.%i.dlt", dltdata->ovaluebase, dltdata->part_num++);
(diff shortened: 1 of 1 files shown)


michael-methner (approved):
Looks good to me. Thanks for your contribution.

PR #596, opened 2024-02-05
Upstream from BOSCH dlt-daemon

minminlittleshrimp:
cmake: correct typo in setting up evironment

unittest: pgrep should detect the exact pattern

cmake: refactor wrong line continuation

unittest: adapt IPv6 testing
+ Checking NULL before checking IP
+ Add environment variable for default loopback IP
+ Add test for IPv6 in unittest, by using converted IP
+ Add correct IP protocol for testing to dlt_gateway.conf


michael-methner:
Hello @minminlittleshrimp ,
i am fine with the changes but could you please add a proper commit message. It should describe what's in the patch not from whom it is coming.

minminlittleshrimp:
> Hello @minminlittleshrimp , i am fine with the changes but could you please add a proper commit message. It should describe what's in the patch not from whom it is coming.

Hello @michael-methner 
kindly review, I already edit the commit title.
Thank you.

michael-methner (approved):

PR #600, opened 2024-02-12
dlt-receive: Synchronize file's in-core state with storage device

### LINKED ISSUE #599: dlt-receive inconsistency between console payload and output file
We have a pc where we run automated tests for a device which is supposed to print the DLT logs in a console while also saving them in a DLT file. The console output is also saved in a separate file. When trying to analyze some logs, we have noticed that while the logs are available in the console's output, logs can not be found in the DLT file for quite some minutes.

Laximas:
Fix #599 
Add fsync check to ensure consistency between console payload and DLT file

### DIFF
--- a/src/console/dlt-receive.c
+++ b/src/console/dlt-receive.c
@@ -88,6 +88,7 @@
 #include "dlt-control-common.h"
 
 #define DLT_RECEIVE_ECU_ID "RECV"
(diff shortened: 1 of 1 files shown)


Laximas:
Implementation does not fix the issue, will close it for now

PR #602, opened 2024-02-12
Upstream portability and security fixes

johnmcfarlane:
- Avoid buffer underrun on readlink failure.
- Don't return int from void function.
- Don't include <sys/syslog.h> which appears to be
  an alternative path for <syslog.h>.


michael-methner (approved):
lgtm. thanks for your contribution.

PR #618, opened 2024-04-05
dlt_multiple_files: Fix missing string null-terminator

asimiklit:
The `newest` string may miss null-terminator, I assume it was supposed to be like this:
```
strncpy(newest, tmp_new, NAME_MAX);
newest[NAME_MAX] = '\0';
```

Note: the link in topic: `Please make sure to follow the [Rules for commit messages](https://at.projects.covesa.org/wiki/display/PROJ/Rules+for+Commit+Messages` does not work for me so commit message maybe incorrect. Error: `We can’t connect to the server at at.projects.covesa.org.`

### DIFF
--- a/src/shared/dlt_multiple_files.c
+++ b/src/shared/dlt_multiple_files.c
@@ -100,7 +100,7 @@ unsigned int multiple_files_buffer_storage_dir_info(const char *path, const char
 
         if ((tmp_new != NULL) && (strlen(tmp_new) < NAME_MAX)) {
             strncpy(newest, tmp_new, NAME_MAX);
-            oldest[NAME_MAX] = '\0';
+            newest[NAME_MAX] = '\0';
         } else if ((tmp_new != NULL) && (strlen(tmp_new) >=  NAME_MAX)) {
             printf("length mismatch of file %s\n", tmp_new);
         }


asimiklit:
But since conditions `(strlen(tmp_new) < NAME_MAX)` and `(strlen(tmp_old) < NAME_MAX)` are present the '\0' should be added by `strncpy` itself:

https://pubs.opengroup.org/onlinepubs/7908799/xsh/strncpy.html
```
If the array pointed to by s2 is a string that is shorter than n bytes, null bytes are appended to the copy in the array pointed to by s1, until n bytes in all are written. 
```
So `newest[NAME_MAX] = '\0'` and `oldest[NAME_MAX] = '\0'` were added as insurance and are redundant.

minminlittleshrimp (changes requested):
Add my review: kindly update your patch.

minminlittleshrimp:
> > Hello @asimiklit I am sorry that I would not take any further action here since your patch is not fully meaningful at all 😢 Maybe you forget to push your latest patch from local? Please kindly check and response. Thank you a lots
> 
> In master you can find such a code:
> 
> ```
>             strncpy(newest, tmp_new, NAME_MAX);
>             oldest[NAME_MAX] = '\0';
> ```
> 
> that looks strange for me since here we are updating `newest` and than add terminator to `oldest` as an insurance but I would expect it to be added to `newest` not `oldest`:
> 
> ```
>             strncpy(newest, tmp_new, NAME_MAX);
>             newest[NAME_MAX] = '\0';
> ```
> 
> But if it is still `not fully meaningful at all` then just close it :)

Thank you for your detail explanation, I understand the issue now 😁 
May the author mistyping here, lol.
Kindly pardon my ignorance.
Thank you for your report and, merged!

minminlittleshrimp (approved):

PR #619, opened 2024-04-09
watchdog: fix another possible watchdog timeout

alexmohr:
dlt_daemon_process_user_message_log prior to this
commit was running without limit, meaning when a lot of messages have to be transmitted dlt-daemon might
be killed by systemd watchdog due to being busy
with sending and not updating the watchdog anymore. This commit introduces a timeout in dlt_daemon_process_user_message_log which allows this function to run at most watchdog_trigger_interval seconds.

This MR extends the logic introduced in https://github.com/COVESA/dlt-daemon/pull/595 as the original commit still allowed systemd to kill dlt-daemon

To test we need the following things
* dlt-daemon started via systemd with enabled watchdog
* a lot of logging going into the daemon

Excecute the script below on your test system.
It spawns a lot of dlt-receive proceses, the goal is for dlt-daemon to survive.

```
end_time=$((SECONDS+180))

while [ $SECONDS -lt $end_time ]; do 
  dlt-receive -a 127.1 &
done 

killall -9 dlt-receive

if [[ $(uname) == "Linux" ]]; then
  systemctl status dlt.service
  echo Service should be running for > 3 minutes
else
  pidin -faAt | grep dlt-daemon | grep -v grep
  echo Service should be running at least since system boot
fi

```

### DIFF
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -3478,7 +3478,22 @@ int dlt_daemon_process_user_message_log(DltDaemon *daemon,
         /* Not enough bytes received to remove*/
         return DLT_DAEMON_ERROR_UNKNOWN;
 
+#ifdef DLT_SYSTEMD_WATCHDOG_ENABLE
+    const unsigned int start_time = dlt_uptime();
+#endif
+
     while (1) {
+#ifdef DLT_SYSTEMD_WATCHDOG_ENABLE
+        const unsigned int uptime = dlt_uptime();
+        if ((uptime - start_time) / 10000 > daemon->watchdog_trigger_interval) {
+            dlt_vlog(LOG_WARNING,
+                     "spent already 1 watchdog trigger interval in %s, yielding to process other events.\n", __func__);
+            if (sd_notify(0, "WATCHDOG=1") < 0)
+                dlt_vlog(LOG_CRIT, "Could not reset systemd watchdog from %s\n", __func__);
+            break;
+        }
+#endif
+
         /* get log message from SHM then store into receiver buffer */
         size = dlt_shm_pull(&(daemon_local->dlt_shm),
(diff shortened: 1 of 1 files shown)


minminlittleshrimp:
imho, I am not sure about the fix.
Honestly I have not touched Watchdog feature in dlt yet 😁 
I would love to spend sometime to check from my local some functionality of this option.
Thank you for the patch!

src/daemon/dlt-daemon.c
```diff
@@ -3478,7 +3478,22 @@ int dlt_daemon_process_user_message_log(DltDaemon *daemon,
...
         return DLT_DAEMON_ERROR_UNKNOWN;
 
+#ifdef DLT_SYSTEMD_WATCHDOG_ENABLE
+    const unsigned int start_time = dlt_uptime();
+#endif
+
     while (1) {
+#ifdef DLT_SYSTEMD_WATCHDOG_ENABLE
+        const unsigned int uptime = dlt_uptime();
+        if ((uptime - start_time) / 10000 > daemon->watchdog_trigger_interval) {
```
michael-methner, line 3488: when this situation occurs, the sd_notify is issued with every processed message. Increasing the system load much more.

I see two options here:
1) avoid looping over serveral messages here
2) update the start_time in case the sd_notify was triggered.

@alexmohr : what do you prefer?

alexmohr: @michael-methner  in line 3493 the `while(1)` is canceled via `break`, meaning we don't call sd_notifiy each time. 
The closing braces belong to the `if` in line 3488. 
Therefore updating the start_time is superflous. Or am I missing something?

michael-methner: @alexmohr : No, I was missing the break(); . THank you very much.

michael-methner (approved):

PR #620, opened 2024-04-18
dlt-receive: add reconnect option 

maiananthan:
## Issue
* When dlt-receive is not connected to dlt-daemon, dlt-receive process is getting ended.

## Fix
* Added a option (`-r`) to reconnect dlt-daemon automatically. 
* Added a option value which can be mentioned as milli seconds as user input.

### DIFF
--- a/src/console/dlt-receive.c
+++ b/src/console/dlt-receive.c
@@ -83,6 +83,8 @@
 #   include <limits.h>
 #endif
 #include <inttypes.h>
+#include <stdbool.h>
+#include <sys/poll.h>
 
 #include "dlt_client.h"
 #include "dlt-control-common.h"
@@ -90,6 +92,7 @@
 #define DLT_RECEIVE_ECU_ID "RECV"
 
 DltClient dltclient;
+static bool sig_close_recv = false;
 
 void signal_handler(int signal)
 {
@@ -99,6 +102,7 @@ void signal_handler(int signal)
     case SIGINT:
     case SIGQUIT:
         /* stop main loop */
+        sig_close_recv = true;
         shutdown(dltclient.receiver.fd, SHUT_RD);
         break;
     default:
@@ -119,12 +123,14 @@ typedef struct {
     int vflag;
     int yflag;
     int uflag;
+    int rflag;
     char *ovalue;
     char *ovaluebase; /* ovalue without ".dlt" */
     char *fvalue;       /* filename for space separated filter file (<AppID> <ContextID>) */
     char *jvalue;       /* filename for json filter file */
     char *evalue;
     int bvalue;
+    int rvalue;
     int sendSerialHeaderFlag;
     int resyncSerialHeaderFlag;
     int64_t climit;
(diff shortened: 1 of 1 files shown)


minminlittleshrimp:
Hello @maiananthan 
thank you.
So it's actually not a bug, but a feature/enhancement.
I'm fine with the point to hang the dlt-client like this, but could you kindly explain a little bit about the purpose?
IMO, dlt client should stop immediately when daemon down, since the pipe/socket closed and nothing further to send/flush to client.
Regards

src/console/dlt-receive.c
```diff
@@ -614,15 +626,26 @@ int main(int argc, char *argv[])
...
+            dlt_client_main_loop(&dltclient, &dltdata, dltdata.vflag);
 
-        /* Dlt Client Cleanup */
-        dlt_client_cleanup(&dltclient, dltdata.vflag);
-    }
+            if (dltdata.rflag == 1 && sig_close_recv == false) {
+                // if reconnect flag is enabled, wait for mentioned milli
+                // seconds
+                poll(NULL, 0, dltdata.rvalue);
+                continue;
```
minminlittleshrimp, line 640: Should the loop stop here? If not, the loop goes on forever!

minminlittleshrimp (changes requested):

minminlittleshrimp:
I suggest my fix, which I applied and checked that it works:
```
$ git df
diff --git a/src/console/dlt-receive.c b/src/console/dlt-receive.c
index ad42e42..58bc0b3 100644
--- a/src/console/dlt-receive.c
+++ b/src/console/dlt-receive.c
@@ -637,14 +637,15 @@ int main(int argc, char *argv[])
                 // if reconnect flag is enabled, wait for mentioned milli
                 // seconds
                 poll(NULL, 0, dltdata.rvalue);
-                continue;
-            }
-            else {
-                /* Dlt Client Cleanup */
-                dlt_client_cleanup(&dltclient, dltdata.vflag);
+                //continue;
                 break;
             }
         }
+        else {
+            /* Dlt Client Cleanup */
+            dlt_client_cleanup(&dltclient, dltdata.vflag);
+            break;
+        }
     } while (dltdata.rflag == 1 && poll(NULL, 0, dltdata.rvalue) == 0);
 
     /* dlt-receive cleanup */
```

src/console/dlt-receive.c
```diff
@@ -162,6 +168,7 @@ void usage()
     printf("  -R            Enable resync serial header\n");
     printf("  -y            Serial device mode\n");
     printf("  -u            UDP multicast mode\n");
+    printf("  -r msecs      Reconnect server with milli seconds specified\n");
```
michael-methner, line 171: Please make the wording more precise: "Reconnect to server after milli seconds"

src/console/dlt-receive.c
```diff
@@ -614,15 +626,26 @@ int main(int argc, char *argv[])
...
+                poll(NULL, 0, dltdata.rvalue);
+                continue;
+            }
+            else {
+                /* Dlt Client Cleanup */
+                dlt_client_cleanup(&dltclient, dltdata.vflag);
+                break;
+            }
+        }
+    } while (dltdata.rflag == 1 && poll(NULL, 0, dltdata.rvalue) == 0);
```
michael-methner, line 648: this is somehow duplicated to the poll instruction before. You could use a while(1) instead.

I would prefer sleep() or nanosleep() instead of the poll to implement a sleep function.

minminlittleshrimp:
Hello @maiananthan 
Kindly update the patch for review and merge.
Thank you a lots.

duvanan13:
Duplicated by #640

PR #664, opened 2024-07-16
daemon: refactor IPC usage

duvanan13:
+ Prevent new daemon initialization when using FIFO
+ Unit test gateway IPC changes to UNIX_SOCKET
+ Shared memory is used for multinode testing
+ Resolve the TOCTOU issue

### DIFF
--- a/src/daemon/dlt-daemon.c
+++ b/src/daemon/dlt-daemon.c
@@ -1585,19 +1585,38 @@ static int dlt_daemon_init_fifo(DltDaemonLocal *daemon_local)
 {
     int ret;
     int fd = -1;
+    int dir_fd = -1;
     int fifo_size;
 
     /* open named pipe(FIFO) to receive DLT messages from users */
     umask(0);
 
-    /* Try to delete existing pipe, ignore result of unlink */
+    /* Valid fifo means there is a daemon running, stop init phase of the new */
     const char *tmpFifo = daemon_local->flags.daemonFifoName;
-    unlink(tmpFifo);
+    if (access(tmpFifo, F_OK) == 0) {
+        dlt_vlog(LOG_WARNING, "FIFO user %s is in use (%s)!\n",
+                 tmpFifo, strerror(errno));
+        return -1;
+    }
 
     ret = mkfifo(tmpFifo, S_IRUSR | S_IWUSR | S_IWGRP);
 
     if (ret == -1) {
         dlt_vlog(LOG_WARNING, "FIFO user %s cannot be created (%s)!\n",
+                 tmpFifo, strerror(errno));
+        return -1;
+    } /* if */
+
+    dir_fd = open(dltFifoBaseDir, O_RDONLY);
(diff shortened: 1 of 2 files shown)


minminlittleshrimp:
Thanks @duvanan13 for joining the pair programming section.
I believe now you have more experience with source control using git.
Thank you for your the rework on this issue, let see how Michael comment about this fix.

Hello @michael-methner , could you kindly review this fix?
Thank you

michael-methner:
Hello @duvanan13 , hello @minminlittleshrimp ,
this patch prevents are restart after dlt-daemon was not shutdown gracefully (e.g. due to a segmentation fault). Could modify the check so it checks for a running instance of dlt-daemon instead of the files present?

duvanan13:
> Hello @duvanan13 , hello @minminlittleshrimp , this patch prevents are restart after dlt-daemon was not shutdown gracefully (e.g. due to a segmentation fault). Could modify the check so it checks for a running instance of dlt-daemon instead of the files present?
In the signal_handle function in daemon, we've already implemented the unlink for FIFO. Hence, my point is to add SIGSEGV to the dlt-daemon signal_handler: https://github.com/COVESA/dlt-daemon/blob/358ab0879dd27b40a9ba4c411c86f7f017e989b5/src/daemon/dlt-daemon.c#L2135
Please kindly check my point.

src/daemon/dlt-daemon.c
```diff
@@ -1414,6 +1414,7 @@ int dlt_daemon_local_init_p1(DltDaemon *daemon, DltDaemonLocal *daemon_local, in
     signal(SIGHUP, dlt_daemon_signal_handler);  /* hangup signal */
     signal(SIGQUIT, dlt_daemon_signal_handler);
     signal(SIGINT, dlt_daemon_signal_handler);
+    signal(SIGSEGV, dlt_daemon_signal_handler);
```
michael-methner, line 1417: The handler for SIGSEGV should not be added here. This will cause the application to hang in an endless loop in case a segmentation fault occurs as it will re-execute the same instruction again which caused the segmentation fault (and signal is triggered again)

src/daemon/dlt-daemon.c
```diff
@@ -2150,6 +2162,7 @@ void dlt_daemon_signal_handler(int sig)
         case SIGTERM:
         case SIGINT:
         case SIGQUIT:
+        case SIGSEGV:
```
andreasbank, line 2165: SIGSEGV needs to be handled separately. For example, after calling dlt_daemon_exit_trigger() set the SIG_DFL for SIGSEGV and raise() the same signal (or just let it hit SIGSEGV on its own again).

I really don't advise calling dlt_daemon_exit_trigger() since it calls async-unsafe functions in a signal handler, but since you already break that rule I wont complain.

minminlittleshrimp: It should be removed

Issue #655, opened 2024-07-03
Save file may  cause mem leak


hush-soul:
I tested it on both qnx and linux
use 
dlt-daemon -d -c /etc/dlt.conf
dlt-logstorage-ctrl -c 1 -C  /etc/dlt.conf -p /etc/dltlog
dlt-example-user -d 5 -n 10000 Hello123

![logstorage](https://github.com/COVESA/dlt-daemon/assets/173775795/0e99cf42-78eb-42e4-83d9-ee3fcac61e40)

hush-soul:
Every time it is stored for a period of time, the memory usage will increase.
Even if the storage quantity has reached the upper limit, old files will be deleted and new files will be stored.

qnx
![dlt_leak_1](https://github.com/COVESA/dlt-daemon/assets/173775795/9a46afe6-6849-4497-90c0-63cf8f32393b)
![dlt_leak_2](https://github.com/COVESA/dlt-daemon/assets/173775795/e94d7cab-9f6c-4c0c-a89d-7b389bb5a15f)
![dlt_leak_3](https://github.com/COVESA/dlt-daemon/assets/173775795/17922613-42a5-4961-941a-0abbf47fe8e7)

hush-soul:
linux
![leak_linux](https://github.com/COVESA/dlt-daemon/assets/173775795/f5f87f9c-090f-471a-b27b-71cc4967537f)

hush-soul:
Initially, I used the log interface of ap autorsa for testing. I created 15 processes and sent the log every 20ms. The memory usage increased wirelessly and very quickly.

In order to rule out AP problems, I used dlt-example-user to test, and the same problem also exists. It is currently under investigation. I would like to ask if there have been similar problems before.

hush-soul:
I observed a phenomenon, dlt-daemon stack = 32kb in qnx, so the memory increases by 32kb each time

hush-soul:
I found that this problem occurs when only performing the first and third steps
1.dlt-daemon -d -c /etc/dlt.conf
2.//dlt-logstorage-ctrl -c 1 -C /etc/dlt.conf -p /etc/dltlog
3.dlt-example-user -d 5 -n 10000 Hello123

hush-soul:
I did some testing on Linux using valgrind
Using 15 processes, each process sends a log every 20ms. The results are shown in Figure 4.
[...]

PR #666, opened 2024-07-17
Solve compress lead mem leak

hush-soul:
#655

### DIFF
--- a/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
+++ b/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
@@ -915,11 +915,6 @@ DLT_STATIC void dlt_logstorage_close_file(DltLogStorageFilterConfig *config)
         gzclose(config->gzlog);
         config->gzlog = NULL;
     }
-    
-    if (config->log) {
-        fclose(config->log);
-        config->log = NULL;
-    }
 #endif
     if (config->log) {
         fclose(config->log);


src/offlinelogstorage/dlt_offline_logstorage.c
```diff
@@ -80,14 +80,16 @@ DLT_STATIC void dlt_logstorage_filter_config_free(DltLogStorageFilterConfig *dat
         data->ecuid = NULL;
     }
 
-    if (data->log != NULL)
-        fclose(data->log);
+
```
minminlittleshrimp, line 83: Could you explain a little bit why we move the code here to other lines?
Is there any reasons for closing the gzip first and then close the log?

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
```diff
@@ -912,6 +915,11 @@ DLT_STATIC void dlt_logstorage_close_file(DltLogStorageFilterConfig *config)
         gzclose(config->gzlog);
         config->gzlog = NULL;
     }
+    
+    if (config->log) {
```
michael-methner, line 919: this code is now duplicated (see lines below the #endif" ). Please remove it from the commit as it will not have any effect

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
```diff
@@ -501,6 +503,7 @@ DLT_STATIC void dlt_logstorage_open_log_output_file(DltLogStorageFilterConfig *c
 #ifdef DLT_LOGSTORAGE_USE_GZIP
         dlt_vlog(LOG_DEBUG, "%s: Opening GZIP log file\n", __func__);
         config->gzlog = gzdopen(config->fd, mode);
+        config->log = file;
```
michael-methner, line 506: there should be no need to store config->log here. As per my understanding gzclose() will close the passed fd from gzdopen()

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
```diff
@@ -371,13 +371,15 @@ int dlt_logstorage_storage_dir_info(DltLogStorageUserConfig *file_config,
     }
 
     char* suffix = NULL;
-    for (i = 0; i < cnt; i++) {
-        if (config->gzip_compression) {
+        
+    if (config->gzip_compression) {
```
michael-methner, line 375: I think allocating the suffix here multiple times in the loops was the only cause for the memory leak. Or is there another memleak fixed in this PR?

src/offlinelogstorage/dlt_offline_logstorage.c
hush-soul: Yes, the order of opening is fopen gzipopen, so I think the order of closing is best gzclose, fclose

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
hush-soul: According to my tests, after gzclose, fclose is also needed. This part of the test is included in #655

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
hush-soul: Thank you, I will test this case separately when I have time to see if it matches the guess

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
hush-soul: @michael-methner  
I did a test to remove fclose, and the results are as follows: fclose is necessary, and the currently submitted code is necessary.

add fclose
add fclose

delete fclose
delete_flocse

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
hush-soul: > I think allocating the suffix here multiple times in the loops was the only cause for the memory leak. Or is there another memleak fixed in this PR?

After testing, it was found that there are two main reasons for memory leaks: 1. fclose needs to be added 2.Allocate memory in a loop

PR #668, opened 2024-07-28
Edited mutex lock of dlt_housekeeper_running_mutex

ranoongs:
The original place of mutex lock was wrong because it didn't protect dlt_housekeeper_running
    and there was no mutex lock when sending signal.

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -3827,12 +3827,17 @@ void dlt_user_housekeeperthread_function(void *ptr)
 
     pthread_cleanup_push(dlt_user_cleanup_handler, NULL);
 
+
+    pthread_mutex_lock(&dlt_housekeeper_running_mutex);
+
     // signal dlt thread to be running
     *dlt_housekeeper_running = true;
     signal_status = pthread_cond_signal(&dlt_housekeeper_running_cond);
     if (signal_status != 0) {
         dlt_log(LOG_CRIT, "Housekeeper thread failed to signal running state\n");
     }
+
+    pthread_mutex_unlock(&dlt_housekeeper_running_mutex);
 
     while (in_loop) {
         /* Check for new messages from DLT daemon */
@@ -5052,6 +5057,10 @@ int dlt_start_threads()
     * (spurious wakeup)
     * To protect against this, a while loop with a timeout is added
     * */
+
+    // pthread_cond_timedwait has to be called on a locked mutex
+    pthread_mutex_lock(&dlt_housekeeper_running_mutex);
+
     while (!dlt_housekeeper_running
            && now.tv_sec <= time_to_wait.tv_sec) {
 
@@ -5064,19 +5073,18 @@ int dlt_start_threads()
         single_wait.tv_sec = now.tv_sec;
         single_wait.tv_nsec = now.tv_nsec + 500000000;
 
-        // pthread_cond_timedwait has to be called on a locked mutex
-        pthread_mutex_lock(&dlt_housekeeper_running_mutex);
         signal_status = pthread_cond_timedwait(
             &dlt_housekeeper_running_cond,
             &dlt_housekeeper_running_mutex,
             &single_wait);
-        pthread_mutex_unlock(&dlt_housekeeper_running_mutex);
 
         /* otherwise it might be a spurious wakeup, try again until the time is over */
         if (signal_status == 0) {
             break;
         }
      }
+
+    pthread_mutex_unlock(&dlt_housekeeper_running_mutex);
 
      if (signal_status != 0 && !dlt_housekeeper_running) {
          dlt_log(LOG_CRIT, "Failed to wait for house keeper thread!\n");


ranoongs:
@michael-methner 
Hello, it's my first time to pull request here. 
Is there any rule about PR?

minminlittleshrimp (commented):
Hello @ranoongs 
The variable you trying to lock is an atomic type and it is totally free from race data.
Hence I believe we do not need to lock the variable.
Kindly check this point.

src/lib/dlt_user.c
```diff
@@ -3827,13 +3827,18 @@ void dlt_user_housekeeperthread_function(void *ptr)
 
     pthread_cleanup_push(dlt_user_cleanup_handler, NULL);
 
+
+    pthread_mutex_lock(&dlt_housekeeper_running_mutex);
+
     // signal dlt thread to be running
     *dlt_housekeeper_running = true;
```
minminlittleshrimp, line 3834: Hello @ranoongs
The variable you trying to lock is an atomic type and it is totally free from race data.
Hence I believe we do not need to lock the variable.
Kindly check this point.

ranoongs: Yes, as you said, dlt_housekeeper_running is atomic.

The objective for this commit is to avoid unnecessary pthread_cond_timedwait's timeout.

lti9hc:
Hello @ranoongs,

You can refer to https://github.com/COVESA/dlt-daemon/pull/491 to know the purpose of using dlt_housekeeper_running_mutex
Moreover, 0.5 sec for timeout is maximum value to wait, it will exit immediately when receiving signal by pthread_cond_signal
For me, This PR should be rejected

ranoongs:
@lti9hc 
Hello,
My point is pthread_mutex_lock(&dlt_housekeeper_running_mutex) is dead code because it misses the corner case I explained above. Yes, I know that 0.5 seconds might not be a big deal but I just want to prevent unnecessary waiting. If the pthread_cond_signal is done 'after' getting into the while loop and 'before' pthread_cond_timedwait, it just waits unnecessary 0.5 seconds. Then what's the point of mutex lock?

If you look into manual page of pthread_cond_timedwait(), mutex guards are out side the while loop. All the examplary source codes use mutex lock to cover all the critical sections...
image

If you say 0.5 seconds is not critical, Well... then I have nothing to say. You can just reject it.

lti9hc:
Hello @ranoongs,

Thanks for your response quickly,

I agree with your point " If the pthread_cond_signal is done 'after' getting into the while loop and 'before' pthread_cond_timedwait, it just waits unnecessary 0.5 seconds"

but with your change, what happen if pthread_mutex_lock(&dlt_housekeeper_running_mutex) in dlt_start_threads function will own mutex first? the dlt_user_housekeeperthread_function thread will block until getting mutex
-> it will wait 0.5s by pthread_cond_timedwait

ranoongs:
@lti9hc 
I thank you too for the quick reply.
**what happen if pthread_mutex_lock(&dlt_housekeeper_running_mutex) in dlt_start_threads function will own mutex first?**
 -> It locks the mutex , changes the variable dlt_housekeeper_running to be true, and unlocks the mutex. The while loop has not started so far because the mutex was not free. Now, the while loop checks the condition !dlt_housekeeper_running is false so the while loop does not initiate. This case shows how the mutex guard successfully protects the critical section.

lti9hc:
@ranoongs ,
The case you mention is housekeeperthread thread will owns the mutex first
My case is opposite your case

ranoongs:
@lti9hc 
Oh sorry.
In that case, it locks the mutex, pthread_cond_timedwait() activates, and it unlocks that mutex. Then **after**, pthread_cond_signal() activates, and the pthread_cond_timedwait() is released. This case also shows that mutex guard protected the critical section because the corner case I mentioned is prevented. The pthread_cond_signal() cannot activate between the start of while loop and pthread_cond_timedwait().

lti9hc:
@ranoongs

**the pthread_cond_timedwait() will get timeout**

The work flow is :
pthread_mutex_lock(&dlt_housekeeper_running_mutex);
while (!dlt_housekeeper_running && now.tv_sec <= time_to_wait.tv_sec)
pthread_cond_timedwait()
pthread_mutex_unlock(&dlt_housekeeper_running_mutex);

....
//waiting the mutex release
dlt_housekeeper_running = true;
pthread_cond_signal(&dlt_housekeeper_running_cond);
pthread_mutex_unlock(&dlt_housekeeper_running_mutex);

it may get the log "dlt_log(LOG_CRIT, "Failed to wait for house keeper thread!\n");"
and then stop all thread

ranoongs:
@lti9hc 
No, the work flow is as below...

pthread_mutex_lock(&dlt_housekeeper_running_mutex);
while (!dlt_housekeeper_running && now.tv_sec <= time_to_wait.tv_sec)
pthread_cond_timedwait() **-> it unlocks the mutex**
pthread_mutex_unlock(&dlt_housekeeper_running_mutex);

....
//waiting the mutex release (X) **-> pthread_mutex_lock(&dlt_housekeeper_running_mutex);**
dlt_housekeeper_running = true;
pthread_cond_signal(&dlt_housekeeper_running_cond);
pthread_mutex_unlock(&dlt_housekeeper_running_mutex);


I share the test result. 

image


image


image

lti9hc:
@ranoongs 
Thanks for your explaining
I am done for review

@alexander-mohr could you please take time to have a look at this PR?

lti9hc (approved):

lti9hc:
It looks good and merge now

Thanks you @ranoongs

PR #691, opened 2024-10-28
Fix usage of gzip_compression

averater:
If dlt_daemon is compiled with gzip compression any use of GzipCompression will turn it on. This is due to incorrect check of the flag where it is checked as if it was a boolenan when it is an int where only the value DLT_LOGSTORAGE_GZIP_ON means that it should be enabled.


minminlittleshrimp:
The fix is good. Thank you @averater

minminlittleshrimp (approved):

PR #694, opened 2024-10-28
Enable file compression on file rotation

averater:
Add option to have dlt files compressed on file rotation so the current file stays uncompressed but all other are compressed. This is so all files can be copied without issues as the current file otherwise is likely to be truncated.

### DIFF
--- a/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
+++ b/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
@@ -152,7 +152,7 @@ void dlt_logstorage_log_file_name(char *log_file_name,
     }
 
     dlt_logstorage_concat_logfile_name(log_file_name, ".dlt");
-    if (filter_config->gzip_compression) {
+    if (filter_config->gzip_compression == DLT_LOGSTORAGE_GZIP_ON) {
         dlt_logstorage_concat_logfile_name(log_file_name, ".gz");
     }
 }
@@ -370,15 +370,10 @@ int dlt_logstorage_storage_dir_info(DltLogStorageUserConfig *file_config,
         config->records = NULL;
     }
 
-    char* suffix = NULL;
-    for (i = 0; i < cnt; i++) {
-        if (config->gzip_compression) {
-            suffix = strdup(".dlt.gz");
-        }
-        else {
-            suffix = strdup(".dlt");
-        }
+    char *suffix = ".dlt.gz";
+    char *gzsuffix = ".dlt";
 
+    for (i = 0; i < cnt; i++) {
         int len = 0;
         len = strlen(file_name);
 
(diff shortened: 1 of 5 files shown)


lti9hc:
Hello @averater,
Could you rebase your PR, it is old and I could not checkout the PR such as "Enable file gzip config" commit
Thank you

lti9hc:
Do you want to combine https://github.com/COVESA/dlt-daemon/pull/691 into this PR? or we need to review and merge 691 PR first and then you rebase this PR later

averater:
> Hello @averater, Could you rebase your PR, it is old and I could not checkout the PR such as "Enable file gzip config" commit Thank you

Done

averater:
> Do you want to combine #691 into this PR? or we need to review and merge 691 PR first and then you rebase this PR later

I'm not sure how the review process will go for this commit. If there are only minor changes needed then we can abandon 691. But if there are more things needed or if this takes to much time then I think we can get 691 in first as that should be trivial. And then rebase this one on top of 691 once it is merged (that will be an easy rebase).

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
```diff
@@ -383,10 +384,13 @@ int dlt_logstorage_storage_dir_info(DltLogStorageUserConfig *file_config,
...
                 /* <filename>.dlt or <filename>_<tmsp>.dlt */
                 if ((files[i]->d_name[len] == suffix[0]) ||
+                    (files[i]->d_name[len] == gzsuffix[0]) ||
                     (file_config->logfile_timestamp &&
-                     (files[i]->d_name[len] == file_config->logfile_delimiter))) {
+                     (files[i]->d_name[len] ==
+                      file_config->logfile_delimiter))) {
                     current_idx = 1;
-                } else {
+                }
```
minminlittleshrimp, line 392: Please do not touch the format but focusing on code change, it distracts me a lots during review.
Thanks.

minminlittleshrimp:
Hello @averater 
Conflict and need resolving, please take care and ping me for reviewing later.
Thanks

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
averater: This was done automatically by the commit check

minminlittleshrimp:
Please commit with sign-off. Thanks.

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
```diff
@@ -732,6 +736,29 @@ int dlt_logstorage_open_log_file(DltLogStorageFilterConfig *config,
                 }
             }
 
+#ifdef DLT_LOGSTORAGE_USE_GZIP
+            if (config->gzip_compression == DLT_LOGSTORAGE_GZIP_FILE) {
+                tmp = &config->records;
+                while ((*tmp)->next != NULL) {
+                    int len = strlen((*tmp)->name);
+                    const char *suffix = ".dlt";
```
minminlittleshrimp, line 744: I need more context/info/description about this feature:
1. Is it trying to do a dynamic activate compress mode, e.g. logging .dlt and then immediately switch to .dlt.gz using compress function for file/fd?
2. If it does so, why dont we just apply compress function in native logstorage file rotation, adding new func here seems doing a duplicated task?
3. Minor suggestion: adding macro name FILE cannot directly inform user/dev that it is for file rotation, any better naming scheme here?

averater: 1. The feature it to log normal dlt and only when a file is "full" it compress it. This way the stored files are compressed but the current file is uncompressed. This is for the use case of copying stored files while logging is in progress.
I think the purpose of removing older files and having a max number of files is to not fill storage. Then it would make sense to bunch dlt and gz.dlt together into one count.

2. It is quite different tasks to compress into an open file and to compress an entire file. I thought this way was the easiest and cleanest way.

3. I thought of it but couldn't figure out a better name that is nice and short. Maybe ARCHIVE is better?

PR #695, opened 2024-10-29
Close file after write

averater:
This commit is to avoid keeping the log file open when not needed. That is helpful if it is a gz file as it then can be copied without being truncated. If the file is copied at the very same time as it is being written to it will still be truncated, but with this commit that risk is smaller.


averater:
This commit changes the write flow from:
Close file if open
Open file
Write to file

New flow:
Close file if open
Open file
Write to file
Close file

I am not sure if either is very good. The latter seem better to me but perhaps should the file not be closed at all unless needed.

lti9hc (approved):

PR #634, opened 2024-05-20
[DRAFT] dlt: Fix compilation warning

minminlittleshrimp:
+ Add checking for snprintf when parsing data from config file
+ Multiple files logging: Refactor function to optimize memory handling with new log files creation
+ log-storage: strncpy checking

PR #697, opened 2024-10-29
Fix compiler warnings related to strings

averater:
Fix string truncation and overflow warnings so the compiler flags stringop-overflow and stringop-truncation can be enabled.

This shouldn't change any behaviour nor fix any bugs. It is just to be able to run the compiler with those flags so future bugs can be avoided. However I have added some extra checks for string lengths.

Most changes is to avoid using strncat and instead use strcat. This doesn't make the code safer but removes the false safeness of using strncat wrongly.

### DIFF
--- a/src/console/dlt-passive-node-ctrl.c
+++ b/src/console/dlt-passive-node-ctrl.c
@@ -113,7 +113,7 @@ void set_node_id(char *id)
         exit(-1);
     }
     else {
-        strcpy(g_options.node_id, id);
+        strncpy(g_options.node_id, id, DLT_ID_SIZE);
     }
 }
 
--- a/src/console/dlt-passive-node-ctrl.c
+++ b/src/console/dlt-passive-node-ctrl.c
@@ -113,7 +113,7 @@ void set_node_id(char *id)
         exit(-1);
     }
     else {
-        strncpy(g_options.node_id, id, DLT_ID_SIZE);
+        strcpy(g_options.node_id, id);
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/console/dlt-passive-node-ctrl.c
```diff
@@ -113,7 +113,7 @@ void set_node_id(char *id)
         exit(-1);
     }
     else {
-        strncpy(g_options.node_id, id, DLT_ID_SIZE);
+        strcpy(g_options.node_id, id);
```
michael-methner, line 116: This may overwrite the end of array g_options.node_id by one byte. node_id is not handled as a null terminated string. 
strncpy is required to restrict the operation to 4 bytes. DLT IDs are not Null-terminated in dlt-daemon.

averater: Thank you, I have restored it. Since there is no length check of argv this is indeed needed.

lti9hc: @michael-methner 
Good point for no null-terminated for App and Context ID

minminlittleshrimp:
Same topic is here: https://github.com/COVESA/dlt-daemon/pull/634

minminlittleshrimp:
Huge rework here: https://github.com/COVESA/dlt-daemon/pull/732

minminlittleshrimp:
Close as it is not recommended to use strcpy and strcat in modern C standard.
More warnings will come with string cumbersome handling.

PR #698, opened 2024-10-31
Remove free of string litteral

averater:
To use free on memory not assigned by alloc is undefined behavior and could cause bugs in some systems.

Also move assignment out of for loop.

### DIFF
--- a/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
+++ b/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
@@ -370,15 +370,9 @@ int dlt_logstorage_storage_dir_info(DltLogStorageUserConfig *file_config,
         config->records = NULL;
     }
 
-    char* suffix = NULL;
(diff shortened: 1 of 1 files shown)


src/offlinelogstorage/dlt_offline_logstorage_behavior.c
```diff
@@ -370,15 +370,9 @@ int dlt_logstorage_storage_dir_info(DltLogStorageUserConfig *file_config,
         config->records = NULL;
     }
 
-    char* suffix = NULL;
-    for (i = 0; i < cnt; i++) {
-        if (config->gzip_compression) {
-            suffix = strdup(".dlt.gz");
-        }
-        else {
-            suffix = strdup(".dlt");
```
lti9hc, line 379: Hello @averater,
strdup will allocate new memory, so this variable is freed after using

lti9hc (approved):
After reviewing by Michael
The change is good and needed to merge
Thank you @averater

PR #713, opened 2024-11-27
Check gzlog in dlt_logstorage_sync_to_file

erikstenlund:
Adds check in dlt_logstorage_sync_to_file that gzlog is NULL if the dlt-daemon is compiled with gzip, otherwise that statement will always return true if using gzip-compression on logs.


lti9hc (approved):

andreasbank:
This is a major fix for the offline storage gzip functionality, should be prioritized.

lti9hc:
it seems good and I am merging now
thanks @erikstenlund for contribution

PR #714, opened 2024-12-02
Remove DisableNetwork logspam

averater:
make sure the warning about disable network is only sent if disable network flag is set and only send the warning once instead of for every message.


lti9hc (approved):
it looks good to me

PR #718, opened 2024-12-05
Fix file deletion when rotating and some bugfixes

andreasbank:
- Fix file deletion when rotating
- Correct mutexattr init
- Small code cleanup and improvements
- Properly terminate log messages

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -761,7 +761,7 @@ DltReturnValue dlt_init_common(void)
     }
 
     /* Binary semaphore for threads */
-    if ((pthread_attr_init(&dlt_mutex_attr) != 0) ||
+    if ((pthread_mutexattr_init(&dlt_mutex_attr) != 0) ||
         (pthread_mutexattr_settype(&dlt_mutex_attr, PTHREAD_MUTEX_ERRORCHECK) != 0) ||
         (pthread_mutex_init(&dlt_mutex, &dlt_mutex_attr) != 0)) {
         dlt_user_init_state = INIT_UNITIALIZED;
--- a/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
+++ b/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
@@ -706,31 +706,33 @@ int dlt_logstorage_open_log_file(DltLogStorageFilterConfig *config,
             num_log_files += 1;
 
             /* check if number of log files exceeds configured max value */
-            if (num_log_files > config->num_files) {
-                if (!(config->num_files == 1 && file_config->logfile_optional_counter)) {
-                    /* delete oldest */
-                    DltLogStorageFileList **head = &config->records;
-                    DltLogStorageFileList *n = *head;
-                    memset(absolute_file_path,
-                           0,
-                           sizeof(absolute_file_path) / sizeof(char));
-                    strcat(absolute_file_path, storage_path);
-                    strncat(absolute_file_path, (*head)->name, strlen((*head)->name));
-                    dlt_vlog(LOG_DEBUG,
-                             "%s: Remove '%s' (num_log_files: %d, config->num_files:%d, file_name:%s)\n",
-                             __func__, absolute_file_path, num_log_files,
-                             config->num_files, config->file_name);
-                    if (remove(absolute_file_path) != 0)
-                        dlt_log(LOG_ERR, "Could not remove file\n");
-
-                    free((*head)->name);
-                    (*head)->name = NULL;
-                    *head = n->next;
-                    n->next = NULL;
-                    free(n);
+            while (num_log_files > config->num_files) {
+                if (config->num_files == 1 && file_config->logfile_optional_counter) {
+                    break;
                 }
-            }
 
+                /* delete oldest */
+                DltLogStorageFileList **head = &config->records;
+                DltLogStorageFileList *n = *head;
+                memset(absolute_file_path,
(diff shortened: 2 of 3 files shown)


src/offlinelogstorage/dlt_offline_logstorage_behavior.c
```diff
@@ -601,7 +601,7 @@ int dlt_logstorage_open_log_file(DltLogStorageFilterConfig *config,
             }
             config->working_file_name = strdup((*newest)->name);
         }
-        strncat(absolute_file_path, config->working_file_name, strlen(config->working_file_name));
+        strcat(absolute_file_path, config->working_file_name);
```
lti9hc, line 604: can you explain why we need this change?

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
```diff
@@ -1107,18 +1109,9 @@ int dlt_logstorage_prepare_on_msg(DltLogStorageFilterConfig *config,
...
-                        }
-                    }
-                    else {
-                        if (fsync(fileno(config->log)) != 0) {
-                            if (errno != ENOSYS) {
-                                dlt_vlog(LOG_ERR, "%s: failed to sync log file\n", __func__);
-                            }
+                    if (fsync(config->fd) != 0) {
+                        if (errno != ENOSYS) {
+                            dlt_vlog(LOG_ERR, "%s: failed to sync log file\n", __func__);
```
lti9hc, line 1114: could you explain why it is only for non gzip?

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
andreasbank: strcat() already uses strlen() internally, so the change from strcat() to using strncat()+strlen() was pointless and only bloats the code. This change was introduced in a commit in the previous year when someone was changing strcat everywhere without considering every line separately.

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
andreasbank: It is not only for gzip it is still for both cases, since I removed the gzip check.

Since _**fileno(config->gzlog) is config->fd**_ and **_fileno(config->log) is also config->fd_**, then we can replace them both with a with a single fsync() call, regardless if config->gzip_compression is set or not.
The executed functions are exactly the same as before but the code looks a lot simpler.

minminlittleshrimp:
Close by mistake, please check https://github.com/COVESA/dlt-daemon/pull/732 for upcoming vers.

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
minminlittleshrimp: Check the huge refactor: https://github.com/COVESA/dlt-daemon/pull/732
Thanks

minminlittleshrimp: > It is not only for gzip it is still for both cases, since I removed the gzip check.
> 
> Since _**fileno(config->gzlog) is config->fd**_ and **_fileno(config->log) is also config->fd_**, then we can replace them both with a with a single fsync() call, regardless if config->gzip_compression is set or not. The executed functions are exactly the same as before but the code looks a lot simpler.

Not correct, gzlog is *gzFile, it could be handled like this:
```
if (config->gzlog && gzdirect(*config->gzlog)) {
                            int fd = fileno((FILE *)*config->gzlog);
                            if (fd != -1 && fsync(fd) != 0) {
                                if (errno != ENOSYS) {
                                    dlt_vlog(LOG_ERR, "%s: failed to sync gzip log file\n", __func__);
```

andreasbank: I yield. Even if config->fd is the exact same as fileno(config-gzlog), it does not matter since it is supposed to be abstract and the assumption should not be depended on for functionality. I removed this change.

src/lib/dlt_user.c
```diff
@@ -761,7 +761,7 @@ DltReturnValue dlt_init_common(void)
     }
 
     /* Binary semaphore for threads */
-    if ((pthread_attr_init(&dlt_mutex_attr) != 0) ||
+    if ((pthread_mutexattr_init(&dlt_mutex_attr) != 0) ||
```
minminlittleshrimp, line 764: fix already available. Thanks

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
minminlittleshrimp: Kindly rework using snprintf for path preparation. Modern SoC with powerful capability so it not a matter with overhead using snprintf. Btw, an official rework can be considered here using global var for storing logstorage path instead of handling multiple times in rounds of filters.

minminlittleshrimp:
Sorry @andreasbank ,
I afraid you might need to rebase on my fix with artifact_v4 update, the CI/CD seems not work at all 😢

minminlittleshrimp:
Also I close #732 and split it to small phases, will get back on this for co-operation with you.
Phase 1 can be checked here: https://github.com/COVESA/dlt-daemon/pull/736

minminlittleshrimp:
Hello @andreasbank 
Can you kindly please rework this code change?
Thanks

minminlittleshrimp:
No action, close for now and work on v2.19.x beta

PR #720, opened 2024-12-16
CMakeLists.txt: make DLT_WatchdogSec can be set by user

sandy-lcq:
In my test env, WatchdogSec default value 2 is not enough, manually changed to 3 is ok. This makes dlt.service/dlt-system.service start failed during boot time. So, make DLT_WatchdogSec can be set by user, so user can set them to proper value at build time, then service can start successfully in boot time.


systemd/CMakeLists.txt
```diff
@@ -18,10 +18,14 @@ if(WITH_SYSTEMD)
     set(SYSTEMD_CONFIGURATIONS_FILES_DIR ${SYSTEMD_UNITDIR} )
 
     if(WITH_SYSTEMD_WATCHDOG)
-        set( DLT_WatchdogSec 2 )
+        if(NOT DEFINED DLT_WatchdogSec)
```
Bichdao021195, line 21: Hi @Le-Tin  This change will not impact to our current source code.
Also, It offer flexibility for user to define their custom DLT_WatchdogSec .
So i agree to merge this change.

lti9hc (approved):
thanks for your contribution @sandy-lcq
thanks for your review @Bichdao021195

PR #733, opened 2025-02-21
cmake: fix generation of pkg-config files

aox-ak:
Variable ZLIB_LIBRARY is used during generation of the pkg-config files to fill the linker flags. Setting it to ZLIB::ZLIB makes that cmake target bleed into the .pc files.

Revert ZLIB_LIBRARY to -lz and use the ZLIB::ZLIB target directly in all affected invocations of target_link_libraries().


src/core_dump_handler/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ if(WITH_DLT_COREDUMPHANDLER)
     endif(WITH_CITYHASH)
 
     add_executable(dlt-cdh ${dlt_cdh_SRCS})
-    target_link_libraries(dlt-cdh ${ZLIB_LIBRARY})
+    target_link_libraries(dlt-cdh ZLIB::ZLIB)
```
lti9hc, line 32: Hello @aox-ak ,
Could you share a little bit why don't we use ${ZLIB_LIBRARY} with -zl in target link libraries?

aox-ak: See commit 42454f96cf0a95427e8c78a9194c8494593aa0c2 where the author explains why he wanted to switch to using the cmake target ZLIB::ZLIB instead of the fixed string -lz. Unfortunately just setting ZLIB_LIBRARY to ZLIB::ZLIB as done in that commit doesn't work, as it makes the cmake target ZLIB::ZLIB bleed into the pkg-config file. My proposed change mixes both approaches. It sets ZLIB_LIBRARY to "-lz" as before to get a correct pkg-config file, but then explicitly uses the target ZLIB::ZLIB in all calls to target_link_libraries.

lti9hc: @aox-ak thanks for your useful information

lti9hc (approved):

lti9hc:
Thanks for your contribution @aox-ak

Issue #631, opened 2024-05-14
Message is lost in dlt-daemon.


tesaki:
The message is lost if more than 65536 bytes of data accumulates in the kernel's receive buffer, when the dlt-daemon receives message data.
When the message type is APP_MSG, the dlt-daemon reads up to 65,535 bytes from the kernel's receive buffer. (ref. [dlt_receiver_init_global_buffer() function](https://github.com/COVESA/dlt-daemon/blob/master/src/shared/dlt_common.c#L2145))  
At this time, the last message created from the read data is often incomplete. As a result, this incomplete message is detected as an error and discarded.

The following patch introduced a change to discard incomplete messages:
[commit: dlt-daemon: Handle partial message parsing in receiver buffer](https://github.com/COVESA/dlt-daemon/commit/1737de27825c691e925a7d42abe34311159c48c6)

Reverting this patch resolves this issue.

To confirm this phenomenon, DaemonFIFOSize is set to 1MiB in dlt.conf and the following log sending program is used.
```C
#include <stdio.h>
#include <stdlib.h>
#include <dlt.h>

DLT_DECLARE_CONTEXT(con_exa1);

int main()
{
    DLT_REGISTER_APP("TAPP", "Test App");

    DLT_REGISTER_CONTEXT(con_exa1, "TAPP", "Test Context");

    int i;
    for (i = 0; i < 5000; i++) {
        DLT_LOG(con_exa1, DLT_LOG_INFO, DLT_UINT32(i),
         DLT_STRING("TEST ################################################"));
    }

    DLT_UNREGISTER_CONTEXT(con_exa1);

    DLT_UNREGISTER_APP();
}
```

tesaki:
I initially proposed increasing the dlt-daemon read size, but I realized this was incorrect. Therefore, I have updated the issue description accordingly.

Hello @minminlittleshrimp and @Suprathik-N
I would like to ask for some clarification regarding the patch "commit: dlt-daemon: Handle partial message parsing in receiver buffer":
This patch description states that partial messages may be received by the daemon and should be ignored. However, our investigation (*1) has indicated that this situation does not occur in a Linux environment.

Could you kindly provide details on the environment in which this patch was tested?
Additionally, could you share some background on how this patch was incorporated?

Based on our findings, this issue does not appear to occur in practice, leading me to believe that the patch may not be necessary.

*1: Investigation Findings
In a Linux environment, we confirmed that writev() and write() do not perform partial writes.

[...]

PR #739, opened 2025-03-19
Revert "dlt-daemon: Handle partial message parsing in receiver buffer"

tesaki:
This reverts commit 1737de27825c691e925a7d42abe34311159c48c6.

The original commit aimed to address potential partial message reception. However, this issue is not observed on Linux and QNX systems, which are the primary target platforms for dlt-daemon. Furthermore, the changes introduced by the reverted commit have been found to cause message loss in certain scenarios.
Therefore, to ensure the stability and reliability of dlt-daemon, this commit reverts the problematic changes.

This revert addresses issue #631.


lti9hc (approved):
@tesaki thanks for your contribution

PR #747, opened 2025-05-06
fix file-rotation-failure-in-offline-storage

quzn:
## Description
Fixes a bug where the dlt-daemon fails to rotate log files in offline storage mode due to fd unavailable.

## Issue Manifestation
- During initial storage, when no log files exist in the designated directory, rotation occurs correctly as configured in dlt_logstorage.conf
- Upon device reboot, when historical log files are present in the storage directory, there exists a probability of rotation failure, resulting in indefinite growth of a single file. The failure probability exhibits an inverse correlation with the FileSize parameter configured in the filter.

## Trigger Conditions
- When writing the first log message after device reboot, if the combined size of the most recent historical log file and the current message exceeds the configured file size threshold
### Example scenario:
- Config->FileSize: 1000
- LatestFileSize: 900
- LogMessage: 110

## Root Cause
The file handle acquisition function fails to execute, leaving the file handle with its default invalid value.


minminlittleshrimp (changes requested):
need utest

quzn:
OK, will add utest later. Thanks!

PR #748, opened 2025-05-09
Fix occasional dlt-daemon hanging when client process exits on QNX

DevriesL:
Problem: dlt-daemon occasionally freezes on QNX when client processes exit unexpectedly

Cause: Socket errors return EBADF but only EPIPE is handled, and select() blocks indefinitely despite timeout (likely a QNX-specific issue)

Fix:
* Add EBADF error handling alongside EPIPE to properly close invalid sockets
* Replace select() with poll() for improved timeout handling efficiency in log output functions

### DIFF
--- a/Android.bp
+++ b/Android.bp
@@ -30,8 +30,7 @@ cc_defaults {
         "-DDLT_LIB_USE_UNIX_SOCKET_IPC",
         "-DCONFIGURATION_FILES_DIR=\"/vendor/etc\"",
         "-DDLT_USER_IPC_PATH=\"/dev/socket\"",
-        "-DDLT_WRITEV_TIMEOUT_SEC=1",
-        "-DDLT_WRITEV_TIMEOUT_USEC=0",
+        "-DDLT_WRITEV_TIMEOUT_MS=1000",
     ] + [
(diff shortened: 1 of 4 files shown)


minminlittleshrimp:
Hello @DevriesL 
From QNX7.1 document says:
```
FD_SETSIZE specifies the highest file descriptor number that the fd_set structure supports and must be set to one more than the highest file descriptor number that the fd_set objects use. Its default value is 256.
```
And:
```
By default, the OS allows a maximum of 1000 file descriptors per process, which well exceeds the maximum number of descriptors allowed by the default FD_SETSIZE value. To avoid errors, use poll() instead of select() (which also is more efficient at handling a sparse list of file descriptors). Alternatively, set FD_SETSIZE to the appropriate value before you include <sys/select.h> or <sys/time.h>.
```
Link: https://www.qnx.com/developers/docs/7.1/#com.qnx.doc.neutrino.lib_ref/topic/s/select.html

Good catch, could you kindly provide some validations/checking on your side (QNX), i.e. fail logs when select() is used, or benchmarking? If it is not convenient then we just run CI/CD and merge. Thanks.

PR #772, opened 2025-09-05
Allow delimiter in filenames

averater:
Allow delimiter (default:_) in file names. With this fix it is possile to have filters for logstorage with for example filename ECU and ECU_B.

### DIFF
--- a/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
+++ b/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
@@ -284,8 +284,7 @@ unsigned int dlt_logstorage_get_idx_of_log_file(DltLogStorageUserConfig *file_co
     idx = strtol(sptr, &eptr, 10);
 
     if (idx == 0)
-        dlt_log(LOG_ERR,
-                "Unable to calculate index from log file name. Reset to 001.\n");
+        dlt_log(LOG_DEBUG, "Unable to calculate index from log file name.\n");
 
     return idx;
 }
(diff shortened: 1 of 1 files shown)


averater:
> How this fix can affect the case? Can you give a brief explanation here? How idx plays here?

If there are two filenames "ECU_<idx>.dlt" and "ECU_B_<idx>.dlt". Then when searching for idx of "ECU_<idx>.dlt" it will find "ECU_B_<idx>.dlt" and will try to parse "B_<idx>.dlt" to int. Since "B" is not possible to convert to int it will return 0. This means for "ECU_<idx>.dlt" the file counter will restart every time it checks the directory.

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
```diff
@@ -396,6 +395,9 @@ int dlt_logstorage_storage_dir_info(DltLogStorageUserConfig *file_config,
                 if (files[i]->d_name[len] == file_config->logfile_delimiter) {
                     current_idx = dlt_logstorage_get_idx_of_log_file(file_config, config,
                                                                      files[i]->d_name);
+                    if (current_idx == 0) {
+                        continue;
```
lti9hc, line 399: @averater it is good to check here
could you check further case in dlt_logstorage_open_log_file function for below code
            } else {
                idx = dlt_logstorage_get_idx_of_log_file(file_config, config,
                                                         config->working_file_name);
            }

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
```diff
@@ -284,8 +284,7 @@ unsigned int dlt_logstorage_get_idx_of_log_file(DltLogStorageUserConfig *file_co
     idx = strtol(sptr, &eptr, 10);
 
     if (idx == 0)
-        dlt_log(LOG_ERR,
-                "Unable to calculate index from log file name. Reset to 001.\n");
+        dlt_log(LOG_DEBUG, "Unable to calculate index from log file name.\n");
```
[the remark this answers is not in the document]
averater: ok

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
averater: That call is unnecessary as it already has been done in dlt_logstorage_storage_dir_info. I'll remove it.

PR #775, opened 2025-09-08
Bump minimum cmake required version to 3.10, needed for compatibility…

LocutusOfBorg:
… with new cmake 4


minminlittleshrimp:
there already under review PR

PR #777, opened 2025-10-14
Fix CONFIGURATION_FILES_DIR

jacky309:
This fixes an inconsistency between the folder where configuration files such as "dlt.conf" are installed, and the folder where those files are read from, at runtime.


CMakeLists.txt
```diff
@@ -268,7 +268,7 @@ endif()
 
 set(CONFIGURATION_FILES_DIR ${CMAKE_INSTALL_SYSCONFDIR})
 
-add_definitions(-DCONFIGURATION_FILES_DIR="${CMAKE_INSTALL_PREFIX}/${CONFIGURATION_FILES_DIR}")
+add_definitions(-DCONFIGURATION_FILES_DIR="${CONFIGURATION_FILES_DIR}")
```
minminlittleshrimp: Good catch, but not sufficient to my end.
Please use this hint:
```
if(CMAKE_INSTALL_PREFIX STREQUAL "/usr")            
    set(CONFIGURATION_FILES_DIR "/etc")             
else()                                              
    set(CONFIGURATION_FILES_DIR "${CMAKE_INSTALL_PREFIX}/etc")
endif()

add_definitions(-DCONFIGURATION_FILES_DIR="${CONFIGURATION_FILES_DIR}")
```

minminlittleshrimp (changes requested):

minminlittleshrimp:
No response, close

PR #782, opened 2025-11-05
cmake: Fix build errors and warnings

minminlittleshrimp:
The rework addresses multiple build
errors and warnings encountered under
strict compilation flags (e.g., -Werror).

The following issues were resolved:
- Pointer type errors in gzflush calls are now wrapped in #ifdef DLT_LOGSTORAGE_USE_GZIP
to ensure correct compilation when GZIP support is enabled.
- String truncation and buffer overflow warnings in logstorage path and filename handling are
fixed by using snprintf and proper length checks.
- The inline keyword was removed from the declaration of dlt_user_is_logLevel_enabled to match its definition and resolve unused inline warnings.
- The node_id buffer in passive node control is now properly null-terminated after strncpy to prevent
truncation warnings.
- The argument order for calloc in dlt-example-user.c was corrected to avoid transposed-args warnings.


lti9hc (approved):
please go ahead @minminlittleshrimp

PR #783, opened 2025-11-13
Protocol V2 support

erikberghvolvo:


minminlittleshrimp:
Review plan:
- Split changes into several topics: Core, Protocols, Utils, Doc, etc
- Commits shall base on the split plan (we shall have 5-8 commits only) then squash into 1
- Regression testing for basic functionality
- UTest from pipeline
- Compile warning: many flags enabled for Warning as Error

minminlittleshrimp (commented):
Huge rework needed and review on every LoC shall be performed due to quality standard.

PR #789, opened 2025-12-02
Update std to 17

LocutusOfBorg:


LocutusOfBorg:
Hello, new googletest requires 17, so for distro shipping googletest system wide this is a requirement to do tests...

minminlittleshrimp:
Okay, luckily C/C++ code base of dlt seems not being affected, due to c++17 changelist

https://en.cppreference.com/w/cpp/17.html

PR #795, opened 2025-12-23
[BUG] console: Fix buffer overflow

minminlittleshrimp:
Fix dlt sortbytimestamp buffer Overflow by:
- Avoid set num = end+1 (out of range)
- Avoid using num as index in case of entending

Full report and demo can be checked in Issue 792.
Reported by Shangzhi-Xu.


lti9hc (approved):
it fixed now
thanks for your fix
@minminlittleshrimp

Issue #792, opened 2025-12-18
[Bug] Protential heap-buffer-overflow crash in `dlt-sortbytimestamp.c`


ShangzhiXu:
Hi team! Thanks for your great work!
I think I found a critical vulnerability that leads to a Heap Buffer Overflow in the system.

It locates at the `main` function in `src/console/dlt-sortbytimestamp.c`.

### PoC
First, we need to generate a valid DLT file to pass the initial file checks. You can use the following python script:

```python
import struct

with open("poc.dlt", "wb") as f:
    for i in range(20):
        # Storage Header (DLT\x01 + Timestamp + ECU)
        f.write(b"DLT\x01" + struct.pack("<II", i, 0) + b"ECU1")
        # Standard Header (HTYP=0x00, MCNT, LEN=8)
        f.write(b"\x00" + struct.pack("B", i % 256) + struct.pack(">H", 8))
        # Payload
        f.write(b"PAYL")
```

Then, we run 
`./src/console/dlt-sortbytimestamp -b 2 -e 5 poc.dlt output.dlt`

We can see the following crash (Heap Buffer Overflow):
```cpp
=================================================================
==2832739==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x506000000068 at pc 0x5b2e0bbe06ad bp 0x7ffd42682b10 sp 0x7ffd42682b00
WRITE of size 4 at 0x506000000068 thread T0
    #0 0x5b2e0bbe06ac in main /mnt/data3/weisong/ACT/fuz/targets/dlt-daemon/src/console/dlt-sortbytimestamp.c:414
    #1 0x7a140b229d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #2 0x7a140b229e3f in __libc_start_main_impl ../csu/libc-start.c:392
    #3 0x5b2e0bbe0764 in _start (/mnt/data3/weisong/ACT/fuz/targets/dlt-daemon/build-asan/src/console/dlt-sortbytimestamp+0x4764)
==2832739==ABORTING
```

### Root Cause
1. The memory size is calculated based on the count of messages in the selected range, at line `392-394`
```cpp
message_count = (uint32_t) (1 + end - begin);                                         // 1+5-2 = 4
timestamp_index = (TimestampIndex *) malloc(sizeof(TimestampIndex) * (message_count + 1) // allocated 5 items
```
[...]

PR #807, opened 2025-12-26
Increase allocated memory for timestamp index

ShangzhiXu:
Further fix https://github.com/COVESA/dlt-daemon/issues/792

### DIFF
--- a/src/console/dlt-sortbytimestamp.c
+++ b/src/console/dlt-sortbytimestamp.c
@@ -391,7 +391,7 @@ int main(int argc, char *argv[]) {
 
     message_count = (uint32_t) (1 + end - begin);
 
-    timestamp_index = (TimestampIndex *) malloc(sizeof(TimestampIndex) * (message_count + 1));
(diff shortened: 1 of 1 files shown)


ShangzhiXu:
cheers~ let me double check~

ShangzhiXu:
Hi there, sorry for the late reply, I've double checked and this bug is still reproducable on the current version, and the patch works on my side.

It is actually that a off-by-one occurs due to some logic error,
```cpp
    message_count = (uint32_t) (1 + end - begin); // lets assume message_count = a

    timestamp_index = (TimestampIndex *) malloc(sizeof(TimestampIndex) * (message_count + 1)); // alloc: a + 1 
    ...
    timestamp_index[message_count + 1].num = timestamp_index[0].num; // visit timestamp_index[a + 1]
```
Here we write `timestamp_index[a + 1]`, but the array has only ` a + 1` elements, so the valid indices are `[0 .. a`].
This fix locates `message_count + 2` elements to make the valid index range `[0 .. a + 1]`

PR #820, opened 2026-02-12
Fix build new version

LocutusOfBorg:


tests/gtest_dlt_json_filter.cpp
```diff
@@ -68,6 +68,7 @@ TEST(t_dlt_message_print_ascii_with_json_filter, normal)
     EXPECT_LE(DLT_RETURN_OK, dlt_file_free(&file, 0));
 }
 /* Begin Method: dlt_common::dlt_message_print_ascii with json filter*/
+#if 0
```
minminlittleshrimp, line 71: any better way to fix instead of completely discard?

PR #828, opened 2026-03-17
fix build failures

sandy-lcq:
Fix build failures when -DWITH_SYSTEMD=ON, -DWITH_SYSTEMD_WATCHDOG=ON, -DWITH_SYSTEMD_JOURNAL=ON:
Eg:
src/system/dlt-system-syslog.c:91:39: error: conversion from 'int' to '__uint16_t' {aka 'short unsigned int'} may change value [-Werror=conversion]
   91 |     syslog_addr.sin6_port = htons(opts.Port);
      |                                   ~~~~^~~~~

src/system/dlt-system-logfile.c:74:21: error: conversion from 'size_t' {aka 'long unsigned int'} to 'int' may change value [-Werror=conversion]
   74 |             bytes = fread(buffer, 1, sizeof(buffer) - 1, pFile);
      |                     ^~~~~

src/system/dlt-system-syslog.c:120:22: error: conversion from 'ssize_t' {aka 'long int'} to 'int' may change value [-Werror=conversion]
  120 |     int bytes_read = recvfrom(sock, recv_data, RECV_BUF_SZ, 0,
      |                      ^~~~~~~~

src/system/dlt-system-logfile.c:76:23: error: comparison of unsigned expression in '>= 0' is always true [-Werror=type-limits]
   76 |             if (bytes >= 0)
      |                       ^~

 src/system/dlt-system-filetransfer.c:727:56: error: '__builtin___snprintf_chk' output may be truncated before the last format cha    racter [-Werror=format-truncation=]
 10   727 |                         snprintf(tosend, length, "%s/%s", opts->Directory[j], ie->name);

src/system/dlt-system-filetransfer.c:153: error: declaration of 'len' shadows a previous local [-Werror=shadow]

### DIFF
--- a/src/system/dlt-system-filetransfer.c
+++ b/src/system/dlt-system-filetransfer.c
@@ -105,14 +105,14 @@ char *unique_name(char *src)
             DLT_STRING("dlt-system-filetransfer, creating unique temporary file name."));
     time_t t = time(NULL);
     int ok;
-    uint32_t l = getFileSerialNumber(src, &ok) ^ t;
+    uint32_t l = (uint32_t)(getFileSerialNumber(src, &ok) ^ t);
 
     if (!ok)
         return (char *)NULL;
 
     char *basename_f = basename(src);
     /* Length of ULONG_MAX + 1 */
-    int len = 11 + strlen(basename_f);
+    size_t len = 11 + strlen(basename_f);
 
     if (len > NAME_MAX) {
         DLT_LOG(dltsystem, DLT_LOG_WARN,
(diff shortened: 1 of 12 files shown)


src/system/dlt-system-filetransfer.c
```diff
@@ -712,26 +712,29 @@ int process_files(FiletransferOptions const *opts)
...
                                     DLT_LOG_ERROR,
                                     DLT_STRING(
                                         "dlt-system-filetransfer: Very long path for file transfer. Cancelling transfer! Length is: "),
-                                    DLT_INT(length));
+                                    DLT_INT((int)length));
                             return -1;
                         }
 
                         char *tosend = malloc(length);
+#pragma GCC diagnostic push
```
minminlittleshrimp, line 727: Need proper fix

minminlittleshrimp (changes requested):

src/system/dlt-system-filetransfer.c
sandy-lcq: This is for fixing failure:
error: '__builtin___snprintf_chk' output may be truncated before the last format character [-Werror=format-truncation=]

I have changed length to "size_t length = strlen(opts->Directory[j]) + strlen(ie->name) + 2;" ,   and then snprintf(tosend, length, "%s/%s", opts->Directory[j], ie->name);  As I see, malloc tosend with length is enough.  I don't understand why gcc still reported the failure,  I did not find the proper way,  so just skip it. Any suggestion?  Thanks.

sandy-lcq:
update the pull request title since I add another commit.

minminlittleshrimp:
Please resolve conflict, btw, I not yet satisfy with the gcc push pop to avoid truncation checking, we might need proper fix

PR #834, opened 2026-04-02
fix: cap dlt_buffer_get copy size to max_size to prevent overflow (#829)

### LINKED ISSUE #829: [BUG] dlt_buffer_get allows buffer overflow.
In cases where the block saved in DLT buffer is at least 20% bigger than DLT_DAEMON_RCVBUFSIZE, the dlt_buffer_read_block method will cause a buffer overflow.

aki1770-del:
## Problem

`dlt_buffer_get()` in `src/shared/dlt_common.c` logs a warning when `head.size > max_size` but still passes `head.size` to `dlt_buffer_read_block()`, writing past the end of the caller's buffer — CWE-119 heap/stack overflow depending on the call site.

Reproducer (from issue #829):
```c
// write block 20% larger than DLT_DAEMON_RCVBUFSIZE
unsigned char write_data[DLT_DAEMON_RCVBUFSIZE + 2004] = {0};
static char data[DLT_DAEMON_RCVBUFSIZE] = {0};

dlt_buffer_push(&buf, write_data, sizeof(write_data));
dlt_buffer_get(&buf, (unsigned char *)data, sizeof(data), 0);
// → writes DLT_DAEMON_RCVBUFSIZE + 2004 bytes into a DLT_DAEMON_RCVBUFSIZE buffer
```

## Fix (`src/shared/dlt_common.c`)

Three changes, one function:

1. **Cap the copy**: introduce `copy_size = min(head.size, max_size)` and pass it to `dlt_buffer_read_block()`.

2. **Ring buffer consistency**: when clipping occurs and `delete=1`, advance the ring buffer read pointer by the full `head.size` (not `copy_size`) so subsequent reads start at the correct position.

3. **Return value**: return `copy_size` instead of `head.size` to correctly report the number of bytes written to the caller's buffer.

**+18 / -10 lines, 1 file.**

Fixes #829.

### DIFF
--- a/src/shared/dlt_common.c
+++ b/src/shared/dlt_common.c
@@ -4242,21 +4242,29 @@ int dlt_buffer_get(DltBuffer *buf, unsigned char *data, int max_size, int delete
         return DLT_RETURN_ERROR; /* ERROR */
     }
 
-    /* third check size */
-    if (max_size && (head.size > max_size))
+    /* third check size — cap copy to max_size to prevent buffer overflow (CWE-119) */
+    int copy_size = head.size;
+    if (max_size && (head.size > max_size)) {
         dlt_vlog(LOG_WARNING,
-                 "%s: Buffer: Max size is smaller than read header size. Max size: %d\n",
-                 __func__, max_size);
+                 "%s: Buffer: read header size %d exceeds max_size %d, clipping copy\n",
+                 __func__, head.size, max_size);
+        copy_size = max_size;
+    }
 
     /* nothing to do but data does not fit provided buffer */
 
     if ((data != NULL) && max_size) {
-        /* read data */
-        dlt_buffer_read_block(buf, &read, data, (unsigned int)head.size);
+        /* read data — copy at most copy_size bytes to avoid overflow */
+        dlt_buffer_read_block(buf, &read, data, (unsigned int)copy_size);
 
-        if (delete)
+        if (delete) {
+            /* advance read pointer past full head.size block for ring buffer consistency */
+            int next_read = read + (head.size - copy_size);
+            if ((unsigned int)next_read >= buf->size)
+                next_read -= (int)buf->size;
             /* update buffer pointers */
-            ((int *)(buf->shm))[1] = read; /* set new read pointer */
+            ((int *)(buf->shm))[1] = next_read;
+        }
 
     }
     else if (delete)
@@ -4276,7 +4284,7 @@ int dlt_buffer_get(DltBuffer *buf, unsigned char *data, int max_size, int delete
             dlt_buffer_minimize_size(buf);
     }
 
-    return head.size; /* OK */
+    return copy_size; /* OK */
 }
 
 int dlt_buffer_pull(DltBuffer *buf, unsigned char *data, int max_size)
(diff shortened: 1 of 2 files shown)


src/shared/dlt_common.c
```diff
@@ -4242,21 +4242,29 @@ int dlt_buffer_get(DltBuffer *buf, unsigned char *data, int max_size, int delete
...
 
     if ((data != NULL) && max_size) {
-        /* read data */
-        dlt_buffer_read_block(buf, &read, data, (unsigned int)head.size);
-
-        if (delete)
+        /* read data — copy at most copy_size bytes to avoid overflow */
+        dlt_buffer_read_block(buf, &read, data, (unsigned int)copy_size);
+
+        if (delete) {
```
e-souza5, line 4260: I believe that updating the read pointer based on the real copy size is a mistake since it won't point to the next header.

e-souza5 (changes requested):

aki1770-del:
@e-souza5 — apologies for the very slow reply, and thank you for the review.

You are right about the invariant: records are stored back-to-back as `DltBufferBlockHead` + `head.size` payload (`dlt_buffer_push3`), so on delete the read pointer must advance past the full `head.size` — never the truncated copy size — or the next `dlt_buffer_get()` lands mid-record and fails the header check.

The diff computes exactly that, which may not be obvious from the hunk: after `dlt_buffer_read_block(buf, &read, data, copy_size)` the local `read` has advanced by `copy_size`, so

```
next_read = read + (head.size - copy_size)  =  (position after block head) + head.size
```

i.e. the start of the next record's header — the same computation as the existing skip branch (`else if (delete)`), just reached via the copy. If you see a case where this still misses the next header, please point it out and I'll fix it.

That said, your changes-requested stands on its own merits — the same hunk had two real defects flagged in the automated review comments, and the revision just pushed (8b1e1b2) fixes both:

1. **Negative `max_size`**: `copy_size = max_size` with `max_size < 0` flowed the negative value into `dlt_buffer_read_block()`, where it is cast to `unsigned int` — a huge memcpy. Reproduced before the fix (segfault right after the `clipping copy` warning with `max_size = -1`). Fixed by copying only when `max_size > 0`; `max_size <= 0` keeps the established delete/skip contract and return value (`dlt_buffer_remove` semantics unchanged).

2. **`>=` wrap**: normalizing `next_read == buf->size` to `0` broke the existing convention — `dlt_buffer_read_block()` itself can leave `read == buf->size`, and the emptiness check requires `write == read` exactly. A clipped record ending exactly at the buffer boundary triggered a spurious "SHM should be empty, but is not" reset (reproduced: write=1012 vs read=0 before the fix). Now wraps on `> buf->size` only, matching the skip branch.

The push also adds four tests in `tests/gtest_dlt_common.cpp`: the #829 reproducer (credit to @edgdsj) asserting no byte lands past `max_size`; a two-record test asserting the record after a clipped delete-get comes back intact (your invariant, tested directly); the exact-boundary case; and the negative-`max_size` case. Local `gtest_dlt_common`: 170/170 passing with the fix; the boundary and negative-`max_size` tests fail/crash on the previous head (exit 133 / 139), which is what confirmed both defects were real.

aki1770-del:
Closing this one.

Sorry for the noise.

(The underlying report #829 is still open — I'm only withdrawing my patch, not the bug.)

PR #836, opened 2026-04-03
Create MSYS2.yaml and fix some issues on MSYS2.

pffang:
Add MSYS2 building action to check building issues on MSYS2.

Fix some "error=char-subscripts" for isdigit and isspace functions by adding explicit cast to int,
which is required by C99 standard and later.


minminlittleshrimp (approved):

minminlittleshrimp (approved):

minminlittleshrimp:
Hello @pffang please answer why we set baudrate to unused? I believe it use in switch case block?

minminlittleshrimp:
continue at https://github.com/COVESA/dlt-daemon/pull/905

Issue #778, opened 2025-10-15
OfflineLogStorage stops outputting logs after repeated enable/disable cycles


mse-kondou-hitomi:
# Steps to reproduce:

1. Enable OfflineLogStorage
2.  Output some logs
3. Disable OfflineLogStorage
4. Modify the OfflineLogStorage configuration file (only change the output file name)

Repeat steps 1–4.

Starting from the third iteration, the following log appears in dlt-daemon, and logs are no longer written to the file:

```
dlt_logstorage_prepare_msg_cache: Max size of Logstorage Cache already used. (ApId=[TIME] CtId=[.*])
dlt_logstorage_write: Unable to prepare. Skip filename [1030401715] because maxmimum trial has been reached.
```

# Configuration details

dlt.conf -> OfflineLogStorageCacheSize is comment out.
```
# Maximal used memory for Logstorage Cache in KB (Default: 30000 KB)
# OfflineLogstorageCacheSize = 30000
```

dlt_logstorage.conf -> FileSize is 10485760(10MB).
```
[FILTER1]
LogAppName=TIME
ContextName=.*
EcuID=ECU1
LogLevel=DLT_LOG_INFO
File=dummy
FileSize=10485760
SyncBehavior=ON_DEMAND
NOFiles=10
```

Upon investigation, it was found that the global variable g_logstorage_cache_size, which manages the total cache size of OfflineLogStorage,
is incremented when OfflineLogStorage is enabled but not decremented when it is disabled.
As a result, when enabling it for the third time, the system determines that the cache size limit has been exceeded.

https://github.com/COVESA/dlt-daemon/blob/f5ea20c11717173c91c74331d47fa6ecc9b2f3a9/src/offlinelogstorage/dlt_offline_logstorage_behavior.c#L1364-L1365

1st: 10MB+16byte < 30MB  ->OK
2nd: 10MB+16byte+10MB+16byte < 30MB  ->OK
3rd: 10MB+16byte+10MB+16byte+10MB+16byte > 30MB  ->NG

# Proposed fix

When disabling OfflineLogStorage, subtract the previously added cache size.

src/offlinelogstorage/dlt_offline_logstorage.c
```
DLT_STATIC void dlt_logstorage_filter_config_free(DltLogStorageFilterConfig *data)
{
     ~~~
    if (data->cache != NULL) {
        free(data->cache);
        data->cache = NULL;

+       unsigned int cache_size = 0;
+      /* check for sync_specific_size strategy */
[...]

PR #838, opened 2026-04-03
fix(logstorage): decrement g_logstorage_cache_size on filter config free

aki1770-del:
## Problem

`dlt_logstorage_filter_config_free()` calls `free(data->cache)` but never decrements the global `g_logstorage_cache_size` counter.

The counter is incremented during cache allocation in `dlt_logstorage_create_cache()`:
```c
g_logstorage_cache_size += cache_size + sizeof(DltLogStorageCacheFooter);
```

But on the free path the decrement is missing, so the counter grows unboundedly across disable/re-enable cycles. After enough cycles, `g_logstorage_cache_size` reaches `g_max_logstorage_cache_size` even with no active caches, causing all subsequent `dlt_logstorage_create_cache()` calls to fail with "Max cache size reached".

Reported in #778.

## Fix

Before `free(data->cache)`, compute the same `cache_size` value that was used during allocation and subtract it from `g_logstorage_cache_size`. Clamp to zero to guard against any bookkeeping inconsistency.

```c
if (data->cache != NULL) {
    unsigned int cache_size = 0;
    if (DLT_OFFLINE_LOGSTORAGE_IS_STRATEGY_SET(data->sync,
            DLT_LOGSTORAGE_SYNC_ON_SPECIFIC_SIZE) > 0)
        cache_size = data->specific_size;
    else
        cache_size = data->file_size;
    cache_size += (unsigned int)sizeof(DltLogStorageCacheFooter);
    if (g_logstorage_cache_size >= cache_size)
        g_logstorage_cache_size -= cache_size;
    else
        g_logstorage_cache_size = 0;
    free(data->cache);
    data->cache = NULL;
}
```

## Testing

Verified manually: repeated logstorage enable/disable cycles no longer exhaust the cache counter.

### DIFF
--- a/src/offlinelogstorage/dlt_offline_logstorage.c
+++ b/src/offlinelogstorage/dlt_offline_logstorage.c
@@ -89,6 +89,17 @@ DLT_STATIC void dlt_logstorage_filter_config_free(DltLogStorageFilterConfig *dat
 #endif
 
     if (data->cache != NULL) {
+        unsigned int cache_size = 0;
+        if (DLT_OFFLINE_LOGSTORAGE_IS_STRATEGY_SET(data->sync,
+                DLT_LOGSTORAGE_SYNC_ON_SPECIFIC_SIZE) > 0)
+            cache_size = data->specific_size;
+        else
+            cache_size = data->file_size;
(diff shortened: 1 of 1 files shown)


aki1770-del:
Closing this one.

Sorry for the noise.

Issue #811, opened 2026-01-12
about pthread_setname_np use(a probability bug in  musl C)


cpunchline:
The assignment of thread IDs and the timing issues in thread callback execution;
pthread_setname_np in thread callback, it should use pthread_self;

dlt-daemon/src/lib/dlt_user.c

minminlittleshrimp:
The statement for root cause is not clear to me, can you state a clear view please?

PR #839, opened 2026-04-03
fix: use pthread_self() for thread naming inside thread callback

aki1770-del:
## Problem

Two thread functions set their thread name using a stored `pthread_t` handle:

```c
// In dlt_segmented_nwt_function():
pthread_setname_np(dlt_user.dlt_segmented_nwt_handle, "dlt_segmented");

// In dlt_housekeeperthread_function():
pthread_setname_np(dlt_housekeeperthread_handle, "dlt_housekeeper");
```

On Linux, `pthread_setname_np(3)` with a non-self TID is implemented via `prctl(PR_SET_NAME)` on the target thread, but the stored handle may not be fully committed to the kernel thread table at the moment the call is made (the spawning `pthread_create()` and the thread start racing). Using the stored handle instead of `pthread_self()` inside the thread body is a TOCTOU race.

The POSIX-safe and portable approach is to call `pthread_setname_np(pthread_self(), name)` from within the thread function, which is what this patch does.

Reported in #811.

## Fix

Replace the stored handle with `pthread_self()` at both call sites.

## Testing

Thread names visible in `/proc/<pid>/task/*/comm` and `ps -T` correctly reflect `dlt_segmented` and `dlt_housekeepr` (15-char kernel limit) after the fix.

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -3949,7 +3949,7 @@ void *dlt_user_trace_network_segmented_thread(void *unused)
     /* Unused on purpose. */
     (void)unused;
 #ifdef DLT_USE_PTHREAD_SETNAME_NP
-    if (pthread_setname_np(dlt_user.dlt_segmented_nwt_handle, "dlt_segmented"))
+    if (pthread_setname_np(pthread_self(), "dlt_segmented"))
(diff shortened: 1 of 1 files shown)


aki1770-del:
Closing this one.

Sorry for the noise.

PR #840, opened 2026-04-03
fix(journal): treat missing PRIORITY field as unknown, not Emergency

### LINKED ISSUE #731: Incorrect log level assignment for auditd messages without PRIORITY field
# Environment

- dlt-daemon: (v2.18.10: 0f2d4cfffada6f8448a2cb27995b38eb4271044f )
- systemd: 244.5
- audit: 2.8.5
- linux: 5.4

# Issue

Currently, dlt-daemon assigns the log level "Emergency" to all journal messages from auditd that lack a PRIORITY field.
This behavior originates from the current implementation of [get_journal_msg()](https://github.com/COVESA/dlt-daemon/blob/master/src/system/dlt-system-journal.c#L202-L256), which does not handle cases where the PRIORITY field is missing.

## Example message from auditd
```
Tue 2024-12-24 17:20:49.383000 UTC+9 [s=0272031534e84e679782433f87f4562a;i=56a;b=a5f717e39b1644779bbdff5c09e85249;m=fcc71;t=629ffca66ae5a;x=6093bf34f8a4fa5d]  
    _BOOT_ID=a5f717e39b1644779bbdff5c09e85249  
    _MACHINE_ID=076f208e43f94dab8f210263f82252d0  
    _HOSTNAME=mymachine  
    _UID=0  
    _TRANSPORT=audit  
    _AUDIT_TYPE=1130  
    SYSLOG_FACILITY=4  
    SYSLOG_IDENTIFIER=audit  
    _AUDIT_TYPE_NAME=SERVICE_START  
    _PID=1  
    _AUDIT_LOGINUID=4294967295  
    _AUDIT_SESSION=4294967295  
    _SELINUX_CONTEXT=system_u:system_r:init_t:s0  
    AUDIT_FIELD_COMM=systemd  
    AUDIT_FIELD_EXE=/lib/systemd/systemd  
    AUDIT_FIELD_HOSTNAME=?  
    AUDIT_FIELD_ADDR=?  
    AUDIT_FIELD_TERMINAL=?  
    AUDIT_FIELD_RES=success  
    _SOURCE_REALTIME_TIMESTAMP=1735028449383000  
    _AUDIT_ID=17  
    MESSAGE=SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=systemd-udevd comm="systemd" exe="/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'  
    AUDIT_FIELD_UNIT=systemd-udevd  
```

## Expected Behavior

Since messages from auditd with SYSLOG_FACILITY=4 do not contain a PRIORITY field, dlt-daemon should determine the appropriate log level based on other fields such as _AUDIT_TYPE.

## Possible Improvement

- Modify get_journal_msg() to:
  - Check SYSLOG_FACILITY and _AUDIT_TYPE when PRIORITY is missing.
  - Use a fallback mechanism to infer a more accurate log level rather than defaulting to "Emergency".
    - e.g. If _AUDIT_TYPE is `AVC`, messages will be classified to "Error" and if not, messages will be classified to "INFO".

Would appreciate any thoughts or suggestions on how best to handle this!

aki1770-del:
## Problem

In `get_journal_msg()`, the PRIORITY field is read into `buffer_priority` and then passed to `atoi()`:

```c
dlt_system_journal_get(j, buffer_priority, "PRIORITY", sizeof(buffer_priority));
...
systemd_loglevel = atoi(buffer_priority);
```

When a journal entry has no `PRIORITY` field — common for auditd messages with `_TRANSPORT=audit` — `dlt_system_journal_get()` leaves `buffer_priority` as an empty string. `atoi("")` returns `0`, which maps to severity level 0 (Emergency/Alert/Critical):

```c
case 0:     /* Emergency */
case 1:     /* Alert */
case 2:     /* Critical */
    loglevel = DLT_LOG_FATAL;
```

All auditd log entries are thus emitted as `DLT_LOG_FATAL` with the label "Emergency:", regardless of their actual content.

Fixes #731.

## Fix

Check `buffer_priority[0] != '\0'` before calling `atoi()`. Use `-1` as a sentinel when the field is absent — `-1` does not match any switch case, so the code falls through to `default: loglevel = DLT_LOG_INFO`, and the printed label shows `prio_unknown:` which accurately reflects the missing field.

```c
systemd_loglevel = (buffer_priority[0] != '\0') ? atoi(buffer_priority) : -1;
```

## Testing

Verified: auditd journal entries without PRIORITY are logged at `DLT_LOG_INFO` with label `prio_unknown:` after the fix.

### DIFF
--- a/src/system/dlt-system-journal.c
+++ b/src/system/dlt-system-journal.c
@@ -252,7 +252,9 @@ void get_journal_msg(sd_journal *j, DltSystemConfiguration *config)
 
         /* map log level on demand */
         loglevel = DLT_LOG_INFO;
-        systemd_loglevel = atoi(buffer_priority);
+        /* Use -1 as sentinel when PRIORITY field is absent (e.g. auditd entries).
+         * atoi("") would return 0 (Emergency), giving a misleadingly high severity. */
+        systemd_loglevel = (buffer_priority[0] != '\0') ? atoi(buffer_priority) : -1;
 
         if (config->Journal.MapLogLevels) {
             /* Map log levels from journal to DLT */


aki1770-del:
Closing this one.

Sorry for the noise.

fizzl:
Explain ontology of error priority.

aki1770-del:
The priority is syslog severity, 0–7: Emergency, Alert, Critical, Error, Warning, Notice, Informational, Debug. dlt-system-journal maps that onto DLT's own ordered scale (FATAL=1 … VERBOSE=6). Both are total orders of severity — there is no "unknown" rung on either.

That's the flaw in what I sent. "Treat missing PRIORITY as unknown" imports a category the ontology doesn't have. What actually happens with a missing field is a mapping-policy question, not an ontology one: buffer_priority is left empty, atoi("") returns 0, and 0 is Emergency → FATAL, so a priority-less entry is logged at the top of the scale.

Whether that's right is a tradeoff, not a correctness fact:
- fail-loud (current): every PRIORITY-less entry — and journald/auditd entries routinely have none — becomes "Emergency:". That's the alert-dilution behind #731.
- fail-quiet (what I was reaching for): map absence low, and you can bury a real emergency that arrived without the field.

Neither is unambiguously correct, and "unknown" isn't one of the options the scale offers. My error was framing a policy choice as if there were a missing ontological value to recover. There isn't.

PR #842, opened 2026-04-03
fix(convert): prevent argv out-of-bounds write and use PID-specific workspace

### LINKED ISSUE #793: [BUG] stack-overflow in `dlt-convert.c`
Hi team, I found another small issue here in `dlt-convert.c`

The code in `main` function extends `argc` based on the number of files in `/tmp/dlt_convert_workspace/` at line 400 
```cpp
argc = optind + (n - 2);
```

However, /tmp is a system-wide shared directory. If a malicious user creates `/tmp/dlt_convert_workspace/` before `dlt-convert` is executed, the code will skip directory creation at line ~352, and the contents of `/tmp/dlt_convert_workspace/` become fully attacker-controlled.

If the attacker places a large number of files in that directory, `n` becomes large and `argc` is increased accordingly.
However, the size of `argv` remains unchanged.

Later, the following loop iterates up to the new, inflated `argc` value:

```cpp
    for (index = optind; index < argc; index++) { // argc is larger than original_argc
            ...
            argv[index] = tmp_filename;  // CRASH, the size of argv is equal to original_argc
        }
```
Leading to the crash.

### PoC
We can create a tar file with
```cpp
mkdir -p /tmp/dlt_convert_workspace/
for i in {1..20}; do touch "/tmp/dlt_convert_workspace/f_$i"; done  // what malicious user can do
touch harmless.dlt
tar -cvf trigger.tar harmless.dlt
```
When we run
```
./src/console/dlt-convert -t trigger.tar
```
and we can see
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==3027575==ERROR: AddressSanitizer: stack-overflow on address 0x7ffd27524000 (pc 0x5a1de8786623 bp 0x7ffd275224f0 sp 0x7ffd2751f200 T0)
    #0 0x5a1de8786623 in main /mnt/data3/weisong/ACT/fuz/targets/dlt-daemon/src/console/dlt-convert.c:409
    #1 0x7e4693429d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #2 0x7e4693429e3f in __libc_start_main_impl ../csu/libc-start.c:392
    #3 0x5a1de8786f94 in _start (/mnt/data3/weisong/ACT/fuz/targets/dlt-daemon/build-asan/src/console/dlt-convert+0x4f94)

SUMMARY: AddressSanitizer: stack-overflow /mnt/data3/weisong/ACT/fuz/targets/dlt-daemon/src/console/dlt-convert.c:409 in main
==3027575==ABORTING
```

### FIX

The memory allocated for `argv` by the OS is fixed and cannot be extended. Writing beyond `argv[original_argc]` corrupts the stack. We can allocate a dynamic array (e.g., char **file_list) on the heap using `malloc` to store the list of files to be processed.

If this makes sense, can I submit a pull request to fix it? : )

aki1770-del:
## Problem

`dlt-convert -t <tarball>` has two security issues:

### 1. argv out-of-bounds write (CWE-131 / stack corruption)

After extracting files to the workspace directory, the code overwrites `argc`:
```c
n = scandir(DLT_CONVERT_WS, &files, NULL, alphasort);
argc = optind + (n - 2);   // argc extended based on files in /tmp
```

A subsequent loop then writes to `argv[index]` using the inflated `argc`:
```c
for (index = optind; index < argc; index++) {
    ...
    argv[index] = tmp_filename;   // index may exceed original argc!
}
```

If a malicious local user pre-populates `/tmp/dlt_convert_workspace/` with extra files before execution, `n` is inflated, `argc` is set beyond the original argv array bounds, and `argv[index]` writes corrupt the stack — leading to a stack overflow as demonstrated with AddressSanitizer in #793.

### 2. Predictable workspace path (CWE-377)

The fixed path `/tmp/dlt_convert_workspace/` is world-writable and predictable. Any local user can pre-create it with arbitrary contents before `dlt-convert` is run, triggering the overflow or injecting unexpected `.dlt` files.

Fixes #793.

## Fix

**For (1):** Introduce a `current_file` pointer. In the processing loop, set it to `tmp_filename` in tarball mode or `argv[index]` otherwise. Remove `argv[index] = tmp_filename` entirely — the original `argv` array is never written to.

**For (2):** Use a PID-specific workspace path `/tmp/dlt_convert_ws_<pid>/` (format string `DLT_CONVERT_WS_FMT`). The path is unique per invocation and not guessable by other local processes.

## Testing

- Verified with the PoC from #793 (pre-populated `/tmp/dlt_convert_ws_*/`) — no stack overflow occurs
- Normal `-t` operation with a valid tar file continues to work correctly

### DIFF
--- a/src/console/dlt-convert.c
+++ b/src/console/dlt-convert.c
@@ -93,7 +93,7 @@
 #define COMMAND_SIZE        1024    /* Size of command */
 #define FILENAME_SIZE       1024    /* Size of filename */
 #define DLT_EXTENSION       "dlt"
-#define DLT_CONVERT_WS      "/tmp/dlt_convert_workspace/"
+#define DLT_CONVERT_WS_FMT  "/tmp/dlt_convert_ws_%d/"  /* %d replaced with PID */
 
 /**
  * Print usage information of tool.
@@ -225,6 +225,9 @@ int main(int argc, char *argv[])
 
     /* For handling compressed files */
     char tmp_filename[FILENAME_SIZE] = { 0 };
+    char dlt_convert_ws[FILENAME_SIZE] = { 0 };
+    snprintf(dlt_convert_ws, sizeof(dlt_convert_ws), DLT_CONVERT_WS_FMT, getpid());
+    const char *current_file = NULL;
     struct stat st;
     memset(&st, 0, sizeof(struct stat));
     struct dirent **files = { 0 };
@@ -349,9 +352,9 @@ int main(int argc, char *argv[])
 
     if (tflag) {
         /* Prepare the temp dir to untar compressed files */
-        if (mkdir(DLT_CONVERT_WS, 0700) != 0) {
+        if (mkdir(dlt_convert_ws, 0700) != 0) {
             if (errno != EEXIST) {
-                fprintf(stderr,"ERROR: Cannot create temp dir %s!\n", DLT_CONVERT_WS);
+                fprintf(stderr,"ERROR: Cannot create temp dir %s!\n", dlt_convert_ws);
                 if (ovalue) {
                     close(ohandle);
                     ohandle = -1;
@@ -361,9 +364,9 @@ int main(int argc, char *argv[])
         }
         else {
             if (S_ISDIR(st.st_mode))
-                empty_dir(DLT_CONVERT_WS);
+                empty_dir(dlt_convert_ws);
             else
-                fprintf(stderr, "ERROR: %s is not a directory", DLT_CONVERT_WS);
+                fprintf(stderr, "ERROR: %s is not a directory", dlt_convert_ws);
         }
 
         for (index = optind; index < argc; index++) {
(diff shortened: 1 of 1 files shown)


aki1770-del:
Closing this one.

Sorry for the noise.

fizzl:
A subsequent loop then writes to argv[index] using the inflated argc:

Explain how these words relate to eachother?

aki1770-del:
"Inflated argc" and "writes to argv[index]" connect through the argc reassignment.

In the tar path argc is overwritten away from its real meaning: `argc = optind + (n - 2)`, where n is scandir's count of the workspace directory. argc no longer bounds argv — it now counts unpacked files. If the archive holds more members than there were command-line arguments, argc exceeds argv's real length. That's the "inflated" part.

Then `for (index = optind; index < argc; index++)` runs to that inflated bound, and inside it `argv[index] = tmp_filename` is a write. Once index passes the real argument count, that write lands past the end of argv. So: scandir inflates argc → the loop bound is now larger than argv → `argv[index] = ...` goes out of bounds. That's how the three phrases relate.

The underlying issue is reusing argc/argv as scratch storage for the unpacked-file list; the clean fix is a separate count and array, not patching the bound in place. That part isn't mine to write — I don't know this codebase well enough — and it's your call whether it warrants its own issue.

PR #846, opened 2026-04-06
[FIX] Add macOS compatibility for thread naming and condition variables

skyroom-ant:
- Add __APPLE__ check for pthread_setname_np (macOS uses single-argument version)
- Skip pthread_condattr_setclock on macOS (CLOCK_MONOTONIC not supported)
- Add Darwin case to tests/CMakeLists.txt to avoid linking socket library

Fixes build issues on macOS while maintaining compatibility with Linux and QNX.


minminlittleshrimp:
feel free to re-open and resolve conflict

PR #850, opened 2026-04-11
warnings: Fix clang generated warnings

kraj:
This patch ensures that it can compile with clang-22

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -261,7 +261,11 @@ add_compile_options(
     -Wcast-qual
     -Wno-system-headers
     -Wno-unused-function
-    -Wno-stringop-truncation
+    $<$<AND:$<COMPILE_LANGUAGE:C>,$<C_COMPILER_ID:Clang>>:-Wno-gnu-statement-expression-from-macro-expansion>
+    $<$<AND:$<COMPILE_LANGUAGE:CXX>,$<CXX_COMPILER_ID:Clang>>:-Wno-gnu-statement-expression-from-macro-expansion>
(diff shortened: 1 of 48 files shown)


minminlittleshrimp:
Hi @kraj 
Please kindly provide compile warning log record for review phase.
Regarding build package, is this using Android ndk or Soong sdk for build?

Thanks

CMakeLists.txt
```diff
@@ -261,7 +261,11 @@ add_compile_options(
     -Wcast-qual
     -Wno-system-headers
     -Wno-unused-function
-    -Wno-stringop-truncation
+    $<$<AND:$<COMPILE_LANGUAGE:C>,$<C_COMPILER_ID:Clang>>:-Wno-gnu-statement-expression-from-macro-expansion>
```
minminlittleshrimp, line 264: Need rechecking for bounded cross-toolchain, see:
https://github.com/COVESA/dlt-daemon/blob/075e03f0864f72d67e246fe9f3da62e91a382a80/CMakeLists.txt#L219

(shall pattern like clang/clang++ must be added?)

minminlittleshrimp:
No action, close. And by the way we about to stick to clang 14

kraj:
> Hi @kraj Please kindly provide compile warning log record for review phase. Regarding build package, is this using Android ndk or Soong sdk for build?
> 
> Thanks

this was sometime ago, I am compiling on linux and using clang instead of GCC to do so with yocto/OE. The latest version of clang there is 22.x and this warning is now treated as error.

```
error: function declaration isn’t a prototype [-Werror,-Wstrict-prototypes]
```

PR #852, opened 2026-04-15
dlt-daemon.c: fix wrong len

sandy-lcq:
Set len to 0 makes the application description always empty. One of the failure case:
running "dlt-example-user 'test'",
check "dlt-control -j localhost" will not get application description: APID:LOG-

Expected:
APID:LOG- Test Application for Logging


santhoshsivanhere:
Hello @minminlittleshrimp, I have verified the issue and fix from my end. Could you please and review / approve ?

minminlittleshrimp (approved):
validated, hence merged as it is

PR #859, opened 2026-04-29
dlt-kpi.c, gtest_dlt_common.cpp: Fix FTBFS with gcc-16 and unused var…

LocutusOfBorg:
…iables

https://bugs.debian.org/1133458

(Closes: #1133458)


minminlittleshrimp (approved):
Reviewed, look good to me

PR #860, opened 2026-04-29
Fix various errors in gcc-16 new compiler, being more picky about cas…

LocutusOfBorg:
…ts and unused variables


LocutusOfBorg:
#859

PR #861, opened 2026-05-05
daemon: fix OOB read and parse bug in dlt_daemon_control_get_log_info_v2 (closes #870)

### LINKED ISSUE #870: dlt_daemon_control_get_log_info_v2 never assigns req->apid / req->ctid; downstream lookups get NULL pointers
Function-specific follow-up to #866. Companion to PR #861 (which
addressed the OOB read in the same function but did not address the
parsing).

## Summary

`dlt_daemon_control_get_log_info_v2` in `src/daemon/dlt_daemon_client.c`
(around line 2098) does not actually parse the `apid` / `ctid` fields
out of incoming GET_LOG_INFO V2 requests. `req->apid` and `req->ctid`
remain NULL when passed to `dlt_daemon_application_find_v2` and
`dlt_daemon_context_find_v2`, so any request with a non-empty apid /
ctid is silently treated as a zero-length lookup.

## Evidence

The `DltServiceGetLogInfoRequestV2` struct
(`include/dlt/dlt_common.h:691-696`) declares `apid` and `ctid` as
`char *` pointers, not inline arrays.

In the function:

- Line ~2141: `req = calloc(1, sizeof(...));` — zero-initialises
  `req`, so `req->apid` and `req->ctid` start NULL.
- Line ~2154:
  `dlt_set_id_v2(req->apid, msg->databuffer + db_offset, req->apidlen);`
  `dlt_set_id_v2` early-returns when its destination `id` is NULL
  (`src/shared/dlt_common.c:417`), so this is a no-op. `req->apid`
  remains NULL.
- Line ~2158: same for `req->ctid`.
- Line ~2186:
  `dlt_daemon_application_find_v2(daemon, req->apidlen, req->apid, ...)`
  called with `req->apid == NULL` despite `req->apidlen` being
  non-zero.
- Line ~2199: same for `req->ctid` passed to
  `dlt_daemon_context_find_v2`.

## Reproduction

Send a GET_LOG_INFO V2 request with a non-empty `apid` (e.g. 4 bytes
`"TEST"`) for a registered application. The daemon parses `apidlen`
correctly but never copies the `apid` bytes. The subsequent
`dlt_daemon_application_find_v2(daemon, 4, NULL, ...)` cannot match
the registered application — the response is built as if no
matching app exists.

## Proposed fix

Same shape as #864 for `set_log_level_v2`. When each length is
non-zero, point `req->apid` / `req->ctid` directly into
`msg->databuffer`:

```c
if (req->apidlen > 0) {
    req->apid = (char *)(msg->databuffer + db_offset);
    db_offset = db_offset + (int)req->apidlen;
}
```

Same for `req->ctid`. The message buffer outlives the function call,
so the pointer-into-databuffer trick is lifetime-safe.

I will extend PR #861 with this parse fix as a second commit rather
than open a new PR, since the two fixes touch the same lines and
sharing a PR avoids merge conflicts.

## Related

- #861 — partial fix (OOB read only) for the same function. Will
  be extended to also close this issue.
- #863 — equivalent issue for `set_log_level_v2`.
- #866 — META-Issue covering the V2 control-handler family.

SoundMatt:
## Problem

`dlt_daemon_control_get_log_info_v2` had two compounding bugs, each
addressed by one commit in this PR.

### 1. Out-of-bounds read parsing apidlen / ctidlen

The fixed-size precheck only validated the 11-byte minimum
GET_LOG_INFO V2 request before the function read two attacker-
controlled `uint8_t` length fields (`apidlen`, `ctidlen`) and used
them to advance an offset into `msg->databuffer`. A short message
with non-zero length fields caused the variable-length reads to walk
past the end of `msg->databuffer` (up to ~514 bytes). Reachable from
any client able to talk to the daemon's control socket. Bounded OOB
read; no OOB write.

### 2. Function never actually parses apid / ctid (#870)

`req` is calloc'd, so its `char *apid` / `char *ctid` pointer fields
start NULL. The `dlt_set_id_v2(req->apid, ...)` calls intended to
populate them are no-ops (the helper early-returns when its
destination is NULL). `req->apid` / `req->ctid` therefore stayed
NULL and were passed to `dlt_daemon_application_find_v2` /
`dlt_daemon_context_find_v2` with non-zero corresponding lengths —
turning every non-empty apid / ctid lookup into a zero-length one.
Full analysis in #870.

## Fix

Two commits:

1. **Bounds checks** after each variable-length field is parsed,
   freeing the calloc'd `req` and returning when the message cannot
   satisfy the next read.
2. **Replace the no-op `dlt_set_id_v2(req->apid, ...)` calls with
   conditional pointer assignments** into `msg->databuffer`,
   mirroring the surgical approach taken for `set_log_level_v2` in
   #864 and `unregister_context_v2` in #868.

`req` shares storage with the message buffer for the duration of
this function call, so the pointer-into-databuffer assignments are
lifetime-safe.

## Notes

- Together with #864, #868, and #869, this completes the V2
  control-handler family cleanup enumerated in #866.
- The function has pre-existing leak paths on other early returns
  (the calloc'd `req` is freed only on the normal exit ~line 2532).
  Not addressed here to keep scope focused — could be a separate
  cleanup PR.

Closes #870.
Related: #866 (META).

### DIFF
--- a/src/daemon/dlt_daemon_client.c
+++ b/src/daemon/dlt_daemon_client.c
@@ -2151,10 +2151,30 @@ void dlt_daemon_control_get_log_info_v2(int sock,
     db_offset = db_offset + (int)sizeof(uint8_t);
     memcpy(&(req->apidlen), msg->databuffer + db_offset, sizeof(uint8_t));
     db_offset = db_offset + (int)sizeof(uint8_t);
+
+    /* apidlen and ctidlen are attacker-controlled (uint8_t each, up to 255).
+     * The fixed-size precheck at the top of this function only validates
+     * the empty case (apidlen = ctidlen = 0); before reading the
+     * variable-length apid (apidlen bytes) followed by ctidlen (1 byte),
+     * confirm the message is large enough to contain them. Otherwise
+     * dlt_set_id_v2() reads past msg->databuffer. */
+    if ((db_offset + (int)req->apidlen + (int)sizeof(uint8_t)) > msg->datasize) {
+        free(req);
+        return;
+    }
+
     dlt_set_id_v2(req->apid, (const char *)(msg->databuffer + db_offset), req->apidlen);
     db_offset = db_offset + (int)req->apidlen;
     memcpy(&(req->ctidlen), (const char *)(msg->databuffer + db_offset), sizeof(uint8_t));
     db_offset = db_offset + (int)sizeof(uint8_t);
+
+    /* Same check for ctid (ctidlen bytes) and the trailing com field
+     * (DLT_ID_SIZE bytes). */
+    if ((db_offset + (int)req->ctidlen + DLT_ID_SIZE) > msg->datasize) {
+        free(req);
+        return;
+    }
+
     dlt_set_id_v2(req->ctid, (const char *)(msg->databuffer + db_offset), req->ctidlen);
     db_offset = db_offset + (int)req->ctidlen;
     memcpy((req->com), (const char *)(msg->databuffer + db_offset), DLT_ID_SIZE);
--- a/src/daemon/dlt_daemon_client.c
+++ b/src/daemon/dlt_daemon_client.c
@@ -2145,6 +2145,8 @@ void dlt_daemon_control_get_log_info_v2(int sock,
 
     /* prepare pointer to message request */
     int db_offset = 0;
+    char apid_buf[DLT_V2_ID_SIZE] = {0};
+    char ctid_buf[DLT_V2_ID_SIZE] = {0};
     memcpy(&(req->service_id), msg->databuffer + db_offset, sizeof(uint32_t));
     db_offset = db_offset + (int)sizeof(uint32_t);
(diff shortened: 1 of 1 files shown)


minminlittleshrimp:
Please signoff in your commit

SoundMatt:
Done — added `Signed-off-by` to both commits (force-pushed).

While here I also reworked the apid/ctid parsing to use the shared `dlt_set_id_v2()` helper into local `char[DLT_V2_ID_SIZE]` buffers instead of aliasing the receive buffer, matching the approach taken in #864/#868 (the struct's `apid`/`ctid` are `char *`, so the original `dlt_set_id_v2(req->apid, …)` was a no-op against a NULL destination). The OOB bounds checks are unchanged; the buffers live for the whole function scope and `req` is freed (struct only) before return, so the pointers stay valid for every use.

minminlittleshrimp:
Fix should be handled inside the API dt_set_id logic.

PR #862, opened 2026-05-05
daemon: fix out-of-bounds read in dlt_daemon_control_set_log_level_v2

SoundMatt:
## Problem

Same shape as #861, applied to the parallel SET_LOG_LEVEL V2 handler.

`dlt_daemon_control_set_log_level_v2` validates only the 11-byte fixed
portion of the SET_LOG_LEVEL V2 request before reading two attacker-
controlled `uint8_t` length fields (`apidlen`, `ctidlen`) and using
them to advance a buffer offset. A short message with non-zero length
fields causes `dlt_set_id_v2()`, the trailing 1-byte `log_level` read,
and the final 4-byte `memcpy()` of the `com` field to read past the
end of `msg->databuffer`. Reachable from any client that can talk to
the daemon's control socket. Bounded OOB read; no OOB write.

## Fix

Two bounds checks:
- After `apidlen` is parsed, verify the message contains
  `apidlen + 1` more bytes (apid + the upcoming `ctidlen` byte).
- After `ctidlen` is parsed, verify the message contains
  `ctidlen + 1 + DLT_ID_SIZE` more bytes (ctid + log_level + com).

Unlike `dlt_daemon_control_get_log_info_v2` in #861, the request struct
here lives on the stack, so no allocation cleanup is required on the
early returns.

## Notes

- Natural follow-up to #861. Functionally independent — can be merged
  in either order.
- I noticed but did **not** address a separate, independent bug in this
  function: at lines 3719/3721 the local `apid`/`ctid` pointers
  (declared NULL at 3687/3688) are passed to `dlt_set_id_v2()` while
  still NULL, so the call is a no-op, yet line 3723 (and later
  branches) dereferences them: `apid[apid_length - 1]`. This NULL-
  derefs whenever `apid_length != 0`. Looks like the intent was
  `apid = req.apid;` (pointer assignment). Filing separately since
  it's an orthogonal bug class.

### DIFF
--- a/src/daemon/dlt_daemon_client.c
+++ b/src/daemon/dlt_daemon_client.c
@@ -3682,10 +3682,26 @@ void dlt_daemon_control_set_log_level_v2(int sock,
     offset = offset + (int)sizeof(uint32_t);
     memcpy(&(req.apidlen), msg->databuffer + offset, sizeof(uint8_t));
     offset = offset + (int)sizeof(uint8_t);
+
+    /* apidlen and ctidlen are attacker-controlled (uint8_t each, up to 255).
+     * The fixed-size precheck at the top of this function only validates
+     * the empty case (apidlen = ctidlen = 0); before reading the
+     * variable-length apid (apidlen bytes) followed by ctidlen (1 byte),
+     * confirm the message is large enough to contain them. Otherwise
+     * dlt_set_id_v2() reads past msg->databuffer. */
+    if ((offset + (int)req.apidlen + (int)sizeof(uint8_t)) > msg->datasize)
+        return;
+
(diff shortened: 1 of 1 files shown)


SoundMatt:
⚠️ **Merge-ordering note for maintainers:** this PR and #864 both edit the same lines of `dlt_daemon_control_set_log_level_v2` — #862 adds the `apidlen`/`ctidlen` bounds validation, #864 fixes the apid/ctid parse (now via the `dlt_set_id_v2()` helper). They are designed to stack, but will **conflict textually** if merged independently. Please merge one and rebase the other (either order is fine); the combined result is the fully-validated handler.

minminlittleshrimp:
Fix should be handled inside the API dt_set_id logic.

PR #864, opened 2026-05-05
daemon: parse apid/ctid in dlt_daemon_control_set_log_level_v2

### LINKED ISSUE #863: dlt_daemon_control_set_log_level_v2 never assigns req.apid/req.ctid; silently fails or crashes the daemon
## Summary

`dlt_daemon_control_set_log_level_v2` in `src/daemon/dlt_daemon_client.c`
does not actually parse the `apid` / `ctid` fields out of incoming
SET_LOG_LEVEL V2 control requests. As a result:

- A request with `apidlen == 0` and `ctidlen == 0` is silently no-op'd
  (the function falls through to a path that calls
  `dlt_daemon_context_find_v2` with NULL apid/ctid).
- A request with `apidlen != 0` causes a NULL pointer dereference at
  line ~3723 and crashes the daemon.

The function is reachable from any client that can talk to the daemon's
control socket (dispatched from `dlt_daemon_handle_control_messages_v2`
at line 1273 of the same file).

## Evidence

The `DltServiceSetLogLevelV2` struct
(`include/dlt/dlt_common.h:760-769`) declares `apid` and `ctid` as
**pointers**, not inline arrays:

```c

### PR DESCRIPTION
**Pull request by SoundMatt**
## Problem

`dlt_daemon_control_set_log_level_v2` never actually parsed the
`apid` / `ctid` fields out of incoming SET_LOG_LEVEL V2 requests:

- `req.apid` / `req.ctid` (declared `char *` in
  `DltServiceSetLogLevelV2`) are zero-initialised to NULL.
- The two `dlt_set_id_v2()` calls intended to populate them were
  no-ops because `dlt_set_id_v2` early-returns when its destination is
  NULL.
- Subsequent code at the wildcard / "apid only" / "ctid only" branches
  dereferences the still-NULL local `apid` / `ctid`
  (`apid[apid_length - 1]` etc.) when their length is non-zero,
  crashing the daemon.

Reachable from any client that can talk to the control socket
(dispatched from `dlt_daemon_handle_control_messages_v2`).

Full analysis in #863.

## Fix

When each length field is non-zero, point `req.apid` / `req.ctid`
directly into `msg->databuffer`. Then assign the local `apid` / `ctid`
pointers via plain assignment so the existing branch logic (which keys
off `apid == NULL` / `ctid == NULL`) operates on real values.

`req` shares storage with the message buffer only for the duration of
this function call, so the pointer-into-databuffer trick is lifetime-
safe.

## Notes

- This PR is independent of #862 (the OOB-read fix in the same
  function): they touch overlapping lines but not in conflicting
  ways. Whichever lands first, the other rebases trivially.
- I have not added a unit test because the function is exercised only
  end-to-end with a live daemon and crafted control messages. Happy to
  add one if maintainers can point me at the right harness.

Closes #863.

### DIFF
--- a/src/daemon/dlt_daemon_client.c
+++ b/src/daemon/dlt_daemon_client.c
@@ -3682,12 +3682,23 @@ void dlt_daemon_control_set_log_level_v2(int sock,
     offset = offset + (int)sizeof(uint32_t);
     memcpy(&(req.apidlen), msg->databuffer + offset, sizeof(uint8_t));
     offset = offset + (int)sizeof(uint8_t);
-    dlt_set_id_v2(req.apid, (const char *)(msg->databuffer + offset), req.apidlen);
-    offset = offset + req.apidlen;
+
+    /* Point req.apid into msg->databuffer when apidlen > 0; leave NULL
+     * otherwise. The previous dlt_set_id_v2(req.apid, ...) call was a
+     * no-op because req.apid is a zero-initialised NULL char *.
+     * See #863 for full analysis. */
+    if (req.apidlen > 0) {
+        req.apid = (char *)(msg->databuffer + offset);
+        offset = offset + req.apidlen;
+    }
+
     memcpy(&(req.ctidlen), msg->databuffer + offset, sizeof(uint8_t));
     offset = offset + (int)sizeof(uint8_t);
-    dlt_set_id_v2(req.ctid, (const char *)(msg->databuffer + offset), req.ctidlen);
-    offset = offset + req.ctidlen;
+
+    if (req.ctidlen > 0) {
+        req.ctid = (char *)(msg->databuffer + offset);
+        offset = offset + req.ctidlen;
+    }
     memcpy(&(req.log_level), msg->databuffer + offset, sizeof(uint8_t));
     offset = offset + (int)sizeof(uint8_t);
     memcpy(&(req.com), msg->databuffer + offset, DLT_ID_SIZE);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/daemon/dlt_daemon_client.c [hunk 1]
```diff
@@ -3682,23 +3682,38 @@ void dlt_daemon_control_set_log_level_v2(int sock,
...
     if (daemon_local->flags.enforceContextLLAndTS)
         req.log_level = (uint8_t) getStatus(req.log_level, daemon_local->flags.contextLogLevel);
 
+    /* The earlier dlt_set_id_v2(apid, req.apid, ...) call was a no-op
+     * because the local apid (declared NULL above) was passed as the
+     * destination. With req.apid now correctly assigned, point the
+     * local pointer at it. Same for ctid. */
+    apid = req.apid;
     apid_length = (int8_t) req.apidlen;
-    dlt_set_id_v2(apid, req.apid, req.apidlen);
```
minminlittleshrimp, line 3699: again, why we dont perform a proper fix for the api but replace with raw assign?

src/daemon/dlt_daemon_client.c
```diff
@@ -3682,23 +3682,38 @@ void dlt_daemon_control_set_log_level_v2(int sock,
     offset = offset + (int)sizeof(uint32_t);
     memcpy(&(req.apidlen), msg->databuffer + offset, sizeof(uint8_t));
     offset = offset + (int)sizeof(uint8_t);
-    dlt_set_id_v2(req.apid, (const char *)(msg->databuffer + offset), req.apidlen);
-    offset = offset + req.apidlen;
+
```
minminlittleshrimp, line 3687: Please again try to fix internal logic of dlt_set_id_v2

minminlittleshrimp (changes requested):

SoundMatt:
Thanks @minminlittleshrimp — reworked to use the helper instead of a raw assignment.

`req.apid` / `req.ctid` are `char *` fields that are NULL in the zero-initialised request struct, which is why the original `dlt_set_id_v2(req.apid, …)` call was a silent no-op. I've added two local `char[DLT_V2_ID_SIZE]` buffers and now copy the wire data into them **through `dlt_set_id_v2()`**, then point `req.apid`/`req.ctid` at those buffers. So id initialisation stays inside the shared API (no hand-parsing, no aliasing of the receive buffer).

`dlt_set_id_v2()` itself had no unit-test coverage, so I also added tests in `gtest_dlt_common_v2.cpp` (normal copy, cap/truncation at `DLT_V2_ID_SIZE` with an overflow guard byte, and NULL/zero-length handling). Verified passing under ASan.

Note: the bounds validation of `apidlen`/`ctidlen` against `datasize` is intentionally kept in the sibling OOB-read PR #862 so the two changes stay reviewable independently; merging both yields a fully-validated handler.

SoundMatt:
⚠️ **Merge-ordering note for maintainers:** this PR and #862 both edit the same lines of `dlt_daemon_control_set_log_level_v2` — #864 fixes the apid/ctid parse, #862 adds the `apidlen`/`ctidlen` bounds validation. They are designed to stack, but will **conflict textually** if merged independently. Please merge one and rebase the other (either order is fine); the combined result is the fully-validated handler.

minminlittleshrimp:

PR #865, opened 2026-05-05
lib: fix heap buffer overflow in dlt_with_filename_and_line_number

SoundMatt:
## Problem

`dlt_with_filename_and_line_number()` in `src/lib/dlt_user.c` writes
past the end of a heap allocation when `strlen(fina) > 255`:

```c
dlt_user.filenamelen = (uint8_t)strlen(fina);          // truncates
...
dlt_user.filename = malloc(dlt_user.filenamelen + 1);  // small
strcpy(dlt_user.filename, fina);                       // full
```

`filenamelen` is `uint8_t` (matches the V2 wire-format field width
declared in `include/dlt/dlt_user.h.in:235`), so the cast truncates
`strlen(fina)` modulo 256. The subsequent `malloc` allocates the
truncated length + 1, but `strcpy` copies the entire `fina` including
its real terminator — writing `strlen(fina) - filenamelen` bytes past
the end of the heap chunk.

Reachable in practice via long compile-time `__FILE__` paths
(monorepo-rooted absolute paths regularly exceed 255 chars) and any
application that forwards a user-controlled filename through this
function. Heap-write primitive in the linked-into-everything client
library.

## Fix

Clamp `strlen(fina)` to `UINT8_MAX` before assigning `filenamelen`,
allocate against the clamped length, and copy with `memcpy` + explicit
null terminator instead of `strcpy`. The on-wire field still gets the
maximum the V2 protocol can encode.

## Notes

- I have not added a unit test in this PR. If maintainers can point me
  at a harness for `dlt_user.c` (e.g. via `gtest_dlt_user.cpp`) I'm
  happy to follow up with one.
- Per-call filename length (`dlt_user_param.filenamelen` at line
  6555) appears to use the same `uint8_t` width. Worth a separate
  audit pass to confirm callers can't trigger the same pattern; not
  in this PR's scope.

### DIFF
--- a/src/lib/dlt_user.c
+++ b/src/lib/dlt_user.c
@@ -4673,7 +4673,12 @@ DltReturnValue dlt_with_filename_and_line_number(const char *fina, const int lin
 
     /* Set filename and line number */
     dlt_user.with_filename_and_line_number = 1;
-    dlt_user.filenamelen = (uint8_t)strlen(fina);
+
+    /* filenamelen is uint8_t (matching the V2 wire-format field width).
+     * Clamp to UINT8_MAX so the cast cannot overflow and the heap
+     * allocation cannot be undersized when strlen(fina) > 255. */
+    dlt_user.filenamelen = (strlen(fina) > UINT8_MAX) ? (uint8_t)UINT8_MAX : (uint8_t)strlen(fina);
+
     if (dlt_user.filename != NULL) {
         free(dlt_user.filename);
         dlt_user.filename = NULL;
@@ -4684,7 +4689,8 @@ DltReturnValue dlt_with_filename_and_line_number(const char *fina, const int lin
         dlt_vlog(LOG_ERR, "%s Could not allocate memory for filename", __func__);
         return DLT_RETURN_ERROR;
     }
-    strcpy(dlt_user.filename, fina);
+    strncpy(dlt_user.filename, fina, (size_t)dlt_user.filenamelen);
+    dlt_user.filename[(size_t)dlt_user.filenamelen] = '\0';
 
     dlt_user.linenumber = (uint32_t)linr;
     return DLT_RETURN_OK;


src/lib/dlt_user.c
```diff
@@ -4673,18 +4673,28 @@ DltReturnValue dlt_with_filename_and_line_number(const char *fina, const int lin
 
     /* Set filename and line number */
     dlt_user.with_filename_and_line_number = 1;
-    dlt_user.filenamelen = (uint8_t)strlen(fina);
+
+    /* filenamelen is uint8_t (matching the V2 wire-format field width).
+     * Clamp the source length to UINT8_MAX before truncation so that
+     * the malloc + copy below cannot write past the end of the heap
+     * allocation when strlen(fina) > 255. */
+    size_t fina_len = strlen(fina);
```
minminlittleshrimp: remove reinit of this var, this var is function opt agr

src/lib/dlt_user.c
```diff
@@ -4673,18 +4673,28 @@ DltReturnValue dlt_with_filename_and_line_number(const char *fina, const int lin
...
     }
 
-    dlt_user.filename = (char*)malloc((size_t)dlt_user.filenamelen + 1);
+    dlt_user.filename = (char*)malloc(fina_len + 1);
     if (dlt_user.filename == NULL){
         dlt_vlog(LOG_ERR, "%s Could not allocate memory for filename", __func__);
         return DLT_RETURN_ERROR;
     }
-    strcpy(dlt_user.filename, fina);
+    memcpy(dlt_user.filename, fina, fina_len);
```
minminlittleshrimp: This pattern is not good.
filename is a string, use strncpy for truncation here + 1 null padding in next line is the common pattern.
e.g.

```
strncpy(dlt_user.filename, fina, fina_len-1);
dlt_user.filename[fina_len] = '/0';

```

minminlittleshrimp (changes requested):
The PR is good in Root cause Analysis, please rework with strncpy and then we can merge.
Thank you

SoundMatt:
Thanks for the detailed feedback! Applied both changes in the latest push:

1. **Removed `fina_len` intermediate variable** — `filenamelen` is now set directly via a ternary that clamps at `UINT8_MAX`, eliminating the separate local variable.
2. **Switched to `strncpy`** — replaced `memcpy` + explicit null with `strncpy(dlt_user.filename, fina, (size_t)dlt_user.filenamelen)` followed by the explicit null terminator at `[filenamelen]`, which is the conventional C string-truncation idiom.

minminlittleshrimp (approved):

minminlittleshrimp:

SoundMatt:
Understood — thanks for the context. I can see the fix from this PR has been absorbed into #895 as part of the broader buffer overflow consolidation.

I'll watch #895 for the combined review. If there's anything I can help with there (testing, review comments, etc.) let me know.

Issue #866, opened 2026-05-05
Multiple V2 control-message handlers are broken by char* fields in protocol structs (set_log_level_v2 [#863], set_trace_status_v2, unregister_context_v2, get_log_info_v2)


SoundMatt:
## Summary

Four V2 control-message handlers in `src/daemon/dlt_daemon_client.c`
are broken because the `DltService...V2` protocol structs they use
were defined with `char *` pointer fields (`apid`, `ctid`, `ecid`)
where inline arrays or properly-assigned pointers were intended. The
handlers either:

- cast `msg->databuffer` directly to one of these structs, in which
  case the pointer fields are interpreted as 8 bytes of network input
  (arbitrary-pointer dereference); or
- declare a stack/calloc'd struct, leaving the pointer fields NULL,
  then call `dlt_set_id_v2(struct.apid, ...)` which silently returns
  early because the destination is NULL; or
- never assign the length/pointer fields at all, leaving uninitialised
  stack values to be written to the wire and used as `memcpy` lengths.

Reachable from any client able to send control messages to the
daemon's control socket.

## Affected functions

All four are in `src/daemon/dlt_daemon_client.c`:

### 1. `dlt_daemon_control_set_log_level_v2` (line ~3660)

Already filed as **#863** and patched in **#864**. Stack-allocated
`req`, NULL `req.apid` / `req.ctid` never assigned, NULL deref at
line 3723 when `apidlen != 0`.

### 2. `dlt_daemon_control_set_trace_status_v2` (line ~4020)

**Most severe.** Direct cast of `msg->databuffer` to
`DltServiceSetLogLevelV2 *`:

```c
req = (DltServiceSetLogLevelV2 *)(msg->databuffer);
...
dlt_set_id_v2(apid, req->apid, req->apidlen);
```

`req->apid` is now whatever 8 bytes happen to be at the right offset
in the wire data — an attacker-controlled pointer. `dlt_set_id_v2`
later reads from `req->apid`. **Arbitrary memory read from network
input.** Daemon either crashes (random pointer rarely valid) or leaks
memory at the attacker-supplied address.

Also has the same NULL-local-`apid`/`ctid` deref pattern as #863, and
the same `(int8_t)` cast on `req->apidlen` which produces a negative
array index when `apidlen > 127`.

### 3. `dlt_daemon_control_message_unregister_context_v2` (line ~2807)

Outbound (server-built) control message. Four bugs in one function:

[...]

PR #868, opened 2026-05-05
daemon: fix multi-bug response builder in dlt_daemon_control_message_unregister_context_v2 (closes #867)

### LINKED ISSUE #867: dlt_daemon_control_message_unregister_context_v2 underallocates response, leaks uninitialised stack to wire, and may segfault on every context unregistration
Function-specific follow-up to #866.

## Summary

`dlt_daemon_control_message_unregister_context_v2` in
`src/daemon/dlt_daemon_client.c` (around line 2807) has **four
distinct bugs**, all stemming from the V2-control-handler pattern
documented in #866. The function builds an UNREGISTER_CONTEXT V2
notification message that is broadcast to clients whenever an
application unregisters a context. It is invoked every time a
registered application disconnects (from `dlt-daemon.c:5001`).

## The function as written

```c
int dlt_daemon_control_message_unregister_context_v2(int sock,
                                                  DltDaemon *daemon,
                                                  DltDaemonLocal *daemon_local,
                                                  uint8_t apidlen,
                                                  char *apid,
                                                  uint8_t ctidlen,
                                                  char *ctid,
                                                  char *comid,
                                                  int verbose)
{
    DltMessageV2 msg;
    DltServiceUnregisterContextV2 resp;          // not zero-initialised
    uint8_t contextSize = 0;
    resp.apid = NULL;                            // explicit
    resp.ctid = NULL;                            // explicit
    int offset = 0;
    /* resp.apidlen, resp.ctidlen, resp.service_id, resp.status,
     * resp.comid all UNINITIALISED at this point. */

    PRINT_FUNCTION_VERBOSE(verbose);

    if (daemon == 0)
        return -1;

    if (dlt_message_init_v2(&msg, 0) == DLT_RETURN_ERROR)
        return -1;

    contextSize = (uint8_t)(sizeof(uint32_t) + sizeof(uint8_t) + sizeof(uint8_t)
                          + apidlen + sizeof(uint8_t) + ctidlen + DLT_ID_SIZE);
    msg.datasize = contextSize;
    ...
    msg.databuffer = (uint8_t *)malloc((size_t)msg.datasize);
    msg.databuffersize = msg.datasize;
    ...

    resp.service_id = DLT_SERVICE_ID_UNREGISTER_CONTEXT;
    resp.status = DLT_SERVICE_RESPONSE_OK;
    dlt_set_id_v2(resp.apid, apid, resp.apidlen);   // NULL dest → no-op
    dlt_set_id_v2(resp.ctid, ctid, resp.ctidlen);   // NULL dest → no-op
    dlt_set_id(resp.comid, comid);

    memcpy(msg.databuffer + offset, &(resp.service_id), sizeof(uint32_t));
    offset = offset + (int)sizeof(uint32_t);
    memcpy(msg.databuffer + offset, &(resp.status), sizeof(uint8_t));
    offset = offset + (int)sizeof(uint8_t);
    memcpy(msg.databuffer + offset, &(resp.apidlen), sizeof(uint8_t));   // ← uninit
    offset = offset + (int)sizeof(uint8_t);
    memcpy(msg.databuffer + offset, resp.apid, resp.apidlen);            // ← NULL src
    offset = offset + (int)resp.apidlen;
    memcpy(msg.databuffer + offset, &(resp.ctidlen), sizeof(uint8_t));   // ← uninit
    offset = offset + (int)sizeof(uint8_t);
    memcpy(msg.databuffer + offset, resp.ctid, resp.ctidlen);            // ← NULL src
    offset = offset + (int)resp.ctidlen;
    memcpy(msg.databuffer + offset, resp.comid, DLT_ID_SIZE);
    ...
}
```

### 1. Heap underallocation via `(uint8_t)` truncation (line 2834)

`contextSize` is `uint8_t`. The summed expression evaluates to up to
`4 + 1 + 1 + 255 + 1 + 255 + 4 = 521` bytes, but the cast wraps mod
256. With `apidlen + ctidlen ≳ 245`, `contextSize` is much smaller
than the actual write footprint. `msg.datasize` then underallocates,
and the subsequent `memcpy` chain writes past the heap allocation.

### 2. Stack-memory disclosure to the wire (lines 2861, 2865)

`resp.apidlen` and `resp.ctidlen` are never assigned from the
function's `apidlen` / `ctidlen` parameters. The stack-allocated
`resp` is not zero-initialised. The `memcpy` of `&(resp.apidlen)` /
`&(resp.ctidlen)` into the response thus copies whatever happened to
be on the stack. **Two bytes of recent stack contents leak to every
client receiving the unregistration message, every time an
application disconnects.** Function-pointer addresses, recently-used
secrets, and arbitrary other stack memory could be exposed.

### 3. `memcpy` from NULL source with arbitrary length (lines 2863, 2867)

`resp.apid` is explicitly `NULL`; never reassigned. `resp.apidlen`
is uninitialised stack memory (see #2 above). `memcpy(dst, NULL,
random_length)` is undefined behaviour — typically segfaults if
`random_length` happens to be non-zero. Same for `resp.ctid` /
`resp.ctidlen`. Whether the daemon survives an unregistration depends
on stack-state luck.

### 4. The intended values are never written to the response

The parameters `apidlen`, `apid`, `ctidlen`, `ctid` are passed to the
function but never copied into `resp`. Even when bugs #2 and #3
happen to be benign (uninitialised len = 0, lucky), the wire response
carries `apidlen = <uninit byte>`, `apid = <missing>`,
`ctidlen = <uninit byte>`, `ctid = <missing>`. **Clients receive
malformed unregister notifications that don't identify which
context unregistered.**

## Reproduction

Connect to a running `dlt-daemon` as a client, register an app and
context with apidlen / ctidlen near 4 (typical), then disconnect.
The daemon will execute this function. With apidlen + ctidlen well
under 245 the heap-underalloc bug doesn't bite immediately, but bug
#3 (NULL `memcpy`) will fire if the stack state produces non-zero
`resp.apidlen`. Triggering #1 reliably requires an
attacker-controlled context register with very large apid/ctid (the
length validation upstream of this function would need separate
review to confirm reachability).

## Proposed fix

Assign all `resp` fields explicitly before serialisation:

```c
DltServiceUnregisterContextV2 resp = {0};   /* zero everything by default */

resp.service_id = DLT_SERVICE_ID_UNREGISTER_CONTEXT;
resp.status     = DLT_SERVICE_RESPONSE_OK;
resp.apidlen    = apidlen;
resp.apid       = apid;
resp.ctidlen    = ctidlen;
resp.ctid       = ctid;
dlt_set_id(resp.comid, comid);
```

Replace the `(uint8_t) contextSize` with a wider type that matches
`msg.datasize`:

```c
uint32_t contextSize = sizeof(uint32_t) + sizeof(uint8_t) + sizeof(uint8_t)
                     + apidlen + sizeof(uint8_t) + ctidlen + DLT_ID_SIZE;
msg.datasize = (int32_t)contextSize;
```

The serialisation `memcpy` chain then writes the correct lengths
and the actual apid/ctid bytes (from the `resp.apid` / `resp.ctid`
pointers, which now point at the function's parameters).

PR with this fix is in flight.

SoundMatt:
## Problem

`dlt_daemon_control_message_unregister_context_v2` had four
distinct bugs (full analysis in #867):

1. **Heap underallocation via `(uint8_t)` truncation** — the
   `contextSize` cast wrapped mod 256, undersizing `msg.datasize`
   for the subsequent `memcpy` chain.
2. **Uninitialised-stack disclosure to wire** — `resp.apidlen` and
   `resp.ctidlen` were never assigned, but were `memcpy`'d into the
   response, leaking two bytes of stack contents per unregister
   notification.
3. **`memcpy` from NULL with uninitialised length** — `resp.apid` /
   `resp.ctid` were set to NULL and the subsequent
   `dlt_set_id_v2(resp.apid, ...)` calls no-op'd; the variable-length
   `memcpy`s then ran with a NULL source and a random length.
4. **Wrong wire output** — the function's parameters were never
   copied into `resp`, so even when the daemon survived the call the
   notification carried garbage lengths and zero context bytes.

Reachable on every context unregistration (called from
`dlt-daemon.c:5001`).

## Fix

- Zero-initialise `resp` with `= {0}`.
- Explicitly assign `resp.apidlen` / `resp.apid` / `resp.ctidlen` /
  `resp.ctid` from the function parameters before serialisation.
- Replace the `(uint8_t) contextSize` cast with a `uint32_t`
  accumulator that matches the `int32_t` width of `msg.datasize`.
- Guard the variable-length `memcpy`s with `len > 0 && ptr != NULL`
  so a future caller passing NULL for an empty field can't UB.

## Notes

- This is one of the four V2 control-handler bugs called out in the
  meta-Issue #866. Follow-up PRs for the others
  (`set_trace_status_v2`, `get_log_info_v2` parsing, possibly more)
  will reference #866 too.
- Tested by reading the diff and walking the new control flow; have
  not added a regression test because this code path is exercised
  only end-to-end with a live daemon and a disconnecting application.
  Happy to add one if maintainers can point me at the right harness.

Closes #867.
Related: #866.

### DIFF
--- a/src/daemon/dlt_daemon_client.c
+++ b/src/daemon/dlt_daemon_client.c
@@ -2815,10 +2815,9 @@ int dlt_daemon_control_message_unregister_context_v2(int sock,
                                                   int verbose)
 {
     DltMessageV2 msg;
-    DltServiceUnregisterContextV2 resp;
-    uint8_t contextSize = 0;
-    resp.apid = NULL;
-    resp.ctid = NULL;
+    /* Zero-initialise the entire response struct so apidlen / ctidlen
+     * cannot leak uninitialised stack memory to the wire (see #867 / #866). */
+    DltServiceUnregisterContextV2 resp = {0};
     int offset = 0;
 
     PRINT_FUNCTION_VERBOSE(verbose);
@@ -2830,10 +2829,33 @@ int dlt_daemon_control_message_unregister_context_v2(int sock,
     if (dlt_message_init_v2(&msg, 0) == DLT_RETURN_ERROR)
         return -1;
 
-    /* prepare payload of data */
-    contextSize = (uint8_t)(sizeof(uint32_t) + sizeof(uint8_t) + sizeof(uint8_t) + apidlen + sizeof(uint8_t) + ctidlen + DLT_ID_SIZE);
+    /* Populate response fields from the function parameters. The
+     * DltServiceUnregisterContextV2 struct uses `char *` for apid / ctid;
+     * point them at the caller-supplied buffers (which outlive this call).
+     * Previously these fields were left NULL while dlt_set_id_v2() was
+     * called with the NULL destination, which is a no-op — the apid /
+     * ctid bytes were never copied into the response. resp.apidlen and
+     * resp.ctidlen were also never assigned, so the subsequent memcpy of
+     * &(resp.apidlen) / &(resp.ctidlen) wrote uninitialised stack memory
+     * to the wire. See #866 for context on the broader pattern. */
+    resp.service_id = DLT_SERVICE_ID_UNREGISTER_CONTEXT;
+    resp.status     = DLT_SERVICE_RESPONSE_OK;
+    resp.apidlen    = apidlen;
+    resp.apid       = apid;
+    resp.ctidlen    = ctidlen;
+    resp.ctid       = ctid;
+    dlt_set_id(resp.comid, comid);
 
-    msg.datasize = contextSize;
+    /* prepare payload of data — use a wider type so the size calculation
+     * cannot overflow when apidlen + ctidlen approach their uint8_t maxima.
+     * Previously a (uint8_t) cast wrapped mod 256, underallocating
+     * msg.databuffer and corrupting the heap on the memcpy chain below. */
+    uint32_t contextSize = (uint32_t)sizeof(uint32_t) + (uint32_t)sizeof(uint8_t)
+                         + (uint32_t)sizeof(uint8_t) + (uint32_t)apidlen
+                         + (uint32_t)sizeof(uint8_t) + (uint32_t)ctidlen
+                         + (uint32_t)DLT_ID_SIZE;
+
+    msg.datasize = (int32_t)contextSize;
 
     if (msg.databuffer && (msg.databuffersize < msg.datasize)) {
         free(msg.databuffer);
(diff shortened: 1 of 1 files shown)


src/daemon/dlt_daemon_client.c
```diff
@@ -2815,10 +2815,9 @@ int dlt_daemon_control_message_unregister_context_v2(int sock,
                                                   int verbose)
 {
     DltMessageV2 msg;
-    DltServiceUnregisterContextV2 resp;
-    uint8_t contextSize = 0;
-    resp.apid = NULL;
-    resp.ctid = NULL;
+    /* Zero-initialise the entire response struct so apidlen / ctidlen
+     * cannot leak uninitialised stack memory to the wire (see #867 / #866). */
```
minminlittleshrimp: Remove these redundant comments

src/daemon/dlt_daemon_client.c
```diff
@@ -2830,10 +2829,33 @@ int dlt_daemon_control_message_unregister_context_v2(int sock,
     if (dlt_message_init_v2(&msg, 0) == DLT_RETURN_ERROR)
         return -1;
 
-    /* prepare payload of data */
-    contextSize = (uint8_t)(sizeof(uint32_t) + sizeof(uint8_t) + sizeof(uint8_t) + apidlen + sizeof(uint8_t) + ctidlen + DLT_ID_SIZE);
+    /* Populate response fields from the function parameters. The
```
minminlittleshrimp: Not convince to me, we better use dlt api/helper function in initialization here, if there issue, fix the internal code from helper function, do not parse raw input like this

minminlittleshrimp (commented):
Please recheck my point, and no need for large comments, please just discard comments

For helper function issue if it possible please provide concrete explanation and fix, we should not eliminate the helper function call that built for dlt coding pattern.
Thanks

SoundMatt:
Thanks @minminlittleshrimp — addressed both points:

1. **Removed the redundant comments.**
2. **No more raw input parsing for the ids.** `resp.apid`/`resp.ctid` are `char *` fields, so I now copy the app/context ids through the shared `dlt_set_id_v2()` helper into local `char[DLT_V2_ID_SIZE]` buffers and point `resp.apid`/`resp.ctid` at those — initialisation stays inside the helper API. (The original `dlt_set_id_v2(resp.apid, …)` was a no-op because the destination pointer was NULL.)

The struct is still fully zero-initialised so `apidlen`/`ctidlen` can't leak uninitialised stack to the wire, and the payload size is summed in `uint32_t` so it can't wrap and underallocate the buffer (#867). `dlt_set_id_v2()` now also has unit-test coverage (added in #864).

minminlittleshrimp:

PR #873, opened 2026-05-22
fix(logstorage): replace unbounded strcat with snprintf in storage_dir_info

SoundMatt:
## Summary

`dlt_logstorage_storage_dir_info()` in `dlt_offline_logstorage_behavior.c` builds a file path from three pieces using three chained `strcat()` calls into a fixed-size stack buffer:

```c
char tmpfile[DLT_OFFLINE_LOGSTORAGE_MAX_LOG_FILE_LEN + 1] = { '\0' };  // 121 bytes
if (dir != NULL) {
    strcat(tmpfile, dir);          // up to ~99 chars (from config)
    strcat(tmpfile, "/");          // + 1
}
strcat(tmpfile, files[i]->d_name); // up to NAME_MAX = 255 chars
```

`dir` is derived from a config-bounded field (max `DLT_OFFLINE_LOGSTORAGE_MAX_FILE_NAME_LEN` = 100 chars), so the directory component is bounded. However `files[i]->d_name` comes from `scandir()` and can be up to `NAME_MAX` (255 bytes) on Linux.

`dir(~99) + "/" + d_name(~255) = ~355 bytes` can easily overflow the 121-byte `tmpfile` buffer, causing a stack buffer overflow. The overflow is triggerable whenever the storage directory contains a file whose name exceeds the buffer capacity, which can happen naturally or via a crafted mount-point with long filenames.

## Changes

Replace the three `strcat()` calls with a single `snprintf(..., sizeof(tmpfile), ...)`, which hard-limits the write to the buffer size. Add a `strncpy` for the no-directory branch for consistency.

## Testing

- Paths shorter than `DLT_OFFLINE_LOGSTORAGE_MAX_LOG_FILE_LEN` are copied unchanged.
- Paths longer than the buffer are safely truncated; no stack overflow.

### DIFF
--- a/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
+++ b/src/offlinelogstorage/dlt_offline_logstorage_behavior.c
@@ -430,11 +430,10 @@ int dlt_logstorage_storage_dir_info(DltLogStorageUserConfig *file_config,
 
             char tmpfile[DLT_OFFLINE_LOGSTORAGE_MAX_LOG_FILE_LEN + 1] = { '\0' };
             if (dir != NULL) {
-                /* Append directory path */
-                strcat(tmpfile, dir);
-                strcat(tmpfile, "/");
+                snprintf(tmpfile, sizeof(tmpfile), "%s/%s", dir, files[i]->d_name);
+            } else {
+                strncpy(tmpfile, files[i]->d_name, DLT_OFFLINE_LOGSTORAGE_MAX_LOG_FILE_LEN);
             }
-            strcat(tmpfile, files[i]->d_name);
             (*tmp)->name = strdup(tmpfile);
             (*tmp)->idx = current_idx;
             (*tmp)->next = NULL;


src/daemon/dlt_daemon_client.c
```diff
@@ -2151,12 +2151,43 @@ void dlt_daemon_control_get_log_info_v2(int sock,
     db_offset = db_offset + (int)sizeof(uint8_t);
     memcpy(&(req->apidlen), msg->databuffer + db_offset, sizeof(uint8_t));
     db_offset = db_offset + (int)sizeof(uint8_t);
-    dlt_set_id_v2(req->apid, (const char *)(msg->databuffer + db_offset), req->apidlen);
```
minminlittleshrimp, line 2154: Hello @SoundMatt 
please kindly explain why the official api is removed.
Kindly approach with fixing the official API instead of manually handly app id or ctx id.

minminlittleshrimp (changes requested):
Please kindly check and fix the api logic, not remove it completely.
Thank you

SoundMatt:
Thanks @minminlittleshrimp — I think the confusion was that this branch was accidentally stacked on top of the GET_LOG_INFO V2 commits, so your comment landed on the `get_log_info_v2` apid/ctid handling at `dlt_daemon_client.c:2154` rather than the logstorage change.

I've rebased this PR onto `master` so it now contains **only** the logstorage fix (the single `storage_dir_info` `strcat`→`snprintf` change). No official API is touched here.

The `get_log_info_v2` apid/ctid parsing you flagged has been reworked to use the shared `dlt_set_id_v2()` helper (into local buffers, no raw wire-buffer aliasing) over in its own PR #861 — same approach as #864/#868. Could you re-review the two separately now that they're cleanly split?

src/offlinelogstorage/dlt_offline_logstorage_behavior.c
```diff
@@ -430,11 +430,10 @@ int dlt_logstorage_storage_dir_info(DltLogStorageUserConfig *file_config,
 
             char tmpfile[DLT_OFFLINE_LOGSTORAGE_MAX_LOG_FILE_LEN + 1] = { '\0' };
             if (dir != NULL) {
-                /* Append directory path */
-                strcat(tmpfile, dir);
-                strcat(tmpfile, "/");
+                snprintf(tmpfile, sizeof(tmpfile), "%s/%s", dir, files[i]->d_name);
+            } else {
+                strncpy(tmpfile, files[i]->d_name, DLT_OFFLINE_LOGSTORAGE_MAX_LOG_FILE_LEN);
```
santhoshsivanhere, line 435: When dir == NULL, strncpy(tmpfile, files[i]->d_name, MAX) does not append a '\0' if the source length ≥ MAX. The buffer is only guaranteed to be zero‑initialized initially, but after copying exactly MAX characters, it may lack a terminator.

Fix: snprintf(tmpfile, sizeof(tmpfile), "%s", files[i]->d_name);

minminlittleshrimp:

PR #877, opened 2026-05-26
fix: proper fd lifecycle management to eliminate fdsan crash

tardyp:
fdsan crash is found while integrating dlt-daemon into an android automotive system.
We are not deep experts of the dlt-daemon inner working, but I think that the IA answer is sound.
We could not reproduce the fdsan crash after patch applied.

Root cause: The DLT_CONNECTION_GATEWAY connections share a receiver pointer aliased into DltGatewayConnection.client.receiver. When dlt_connection_destroy() closed the fd, the gateway's client.sock retained the stale fd number. If the kernel reused that fd number (e.g., for a FILE*), subsequent gateway send() calls would trigger Android fdsan abort.

Changes:
- dlt_connection_destroy: Do NOT close fd for GATEWAY connections (they don't own it). Only detach the receiver pointer.
- dlt_gateway_close_connection: New function that properly closes client.sock AND invalidates client.receiver.fd at all disconnect points.
- Deferred destruction: Connections are now marked PENDING_DESTROY instead of being freed immediately. A sweep phase after the event loop iteration safely destroys them, preventing use-after-free when callbacks trigger their own connection's removal.

### DIFF
--- a/src/daemon/dlt_daemon_connection.c
+++ b/src/daemon/dlt_daemon_connection.c
@@ -345,13 +345,38 @@ void *dlt_connection_get_callback(DltConnection *con)
  * Ownership of the connection is given during the registration to
  * the DltEventHandler.
  *
+ * For DLT_CONNECTION_GATEWAY connections, the fd is owned by the gateway's
+ * DltClient structure and must NOT be closed here. The gateway module is
+ * responsible for closing its own socket via dlt_client_cleanup().
+ *
  * @param to_destroy Connection to be destroyed.
  */
 void dlt_connection_destroy(DltConnection *to_destroy)
 {
+    if (to_destroy == NULL)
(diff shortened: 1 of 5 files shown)


minminlittleshrimp:
Hi @tardyp 
good catch.
This can be combined with our latest upstream fix for fds bug in daemon. Shortly say, team has a huge bug fix effort for the plling loop including all fds (unix and inet socket, user fifo or socket handling) but this not yet bounded the multinode-gateway case.
We shall take your patch seriously and provide feedback soon if rework needed, else we would prefer we upstream our fix first and rebase yours upon.
Thanks for the contribution, happy coding
Minh

jbmaillet:
Any news? This PR for a security issue is now 3 months old.

PR #882, opened 2026-06-16
bump cmake version to minimum 3.16.5

rounak-bawankule:
bump cmake version to minimum 3.16.5

- jobs are failing due to older version of cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@
 #######
 
 # Set minimum Cmake version and setup policy behavior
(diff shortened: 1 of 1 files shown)


minminlittleshrimp:
Hi @rounak-bawankule 
So what is the issue we have here?
What problem this patch solved?

rounak-bawankule:
> Hi @rounak-bawankule So what is the issue we have here? What problem this patch solved?

Hi @minminlittleshrimp, this patch was just to test a different version, sorry i should have mentioned in the commit message, closing the PR now

PR #885, opened 2026-06-19
Fix installation of default conf files.

LocutusOfBorg:
CMAKE_INSTALL_SYSCONFDIR points to /etc, and this is where the location of configuration files has to be found.

Otherwise, it becomes /usr/etc, and with configuration files installed elsewhere (/etc), there is a mismatch between where the code tries to find the file, and where the file is actually located, e.g.

INSTALL(FILES dlt-system.conf
        DESTINATION ${CONFIGURATION_FILES_DIR}
        COMPONENT base)

This makes the conf file not being correctly evaluated on startup, because the systemd service script doesn't pass any value for "-c" switch


minminlittleshrimp:
Validated, correct patch fixing

minminlittleshrimp (approved):

LocutusOfBorg:
Thanks!

PR #891, opened 2026-06-26
[Prior1] build: Support both Clang and GCC toolchains

felipek:
-Wno-stringop-truncation is a GCC-only warning option. Passing it
unconditionally broke the Clang build, which rejects it as an unknown
warning option (an error under -Werror). Make it GCC-conditional via a
generator expression.

Additionally add a Clang-only block that relaxes the Clang-specific
diagnostics triggered by the same strict flags (-pedantic, -Wconversion):
intentional GNU extensions (gnu11/gnu++17), "()" vs "(void)" prototypes,
and stricter implicit enum/integer narrowing and signedness checks. The
GCC build keeps its full strict -Werror warning set unchanged.

Verified clean (-Werror, 0 warnings) with Apple Clang 21 and GCC 15.2.0.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -261,8 +261,39 @@ add_compile_options(
     -Wcast-qual
     -Wno-system-headers
     -Wno-unused-function
-    -Wno-stringop-truncation
+    # -Wno-stringop-truncation is a GCC-only warning option; Clang rejects it
+    # as an unknown warning option (which becomes an error under -Werror).
+    $<$<C_COMPILER_ID:GNU>:-Wno-stringop-truncation>
+    $<$<CXX_COMPILER_ID:GNU>:-Wno-stringop-truncation>
 )
+
+# Clang support.
+#
+# The DLT code base is written for and CI-validated with GCC, which keeps the
+# full strict -Werror warning set above. Clang enables several additional or
+# stricter diagnostics for the very same flags (-pedantic, -Wconversion, ...).
+# Most of them are either intentional GNU extensions (the project is compiled
+# as gnu11/gnu++17) or stylistic differences that GCC does not flag. Relax only
+# those Clang-specific categories so Clang can build the project as well, while
+# the GCC build keeps its full strictness unchanged.
+if(CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    add_compile_options(
+        # Empty parameter lists "()" instead of "(void)". Not flagged by the
+        # GCC build; would require touching many public prototypes.
+        $<$<COMPILE_LANGUAGE:C>:-Wno-strict-prototypes>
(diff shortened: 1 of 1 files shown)


felipek:
Hi @alexmohr @minminlittleshrimp please review this as follow-up to https://github.com/COVESA/dlt-daemon/commit/c840f299ca6856e3cdb8b066238f45fd0fdf6f0c - there is a very tiny incompatible change requiring gcc (brew) which is not great since people can use standard Apple/Xcode/clang tools with this minor cmake/make change. I have tested it both with GCC and Clang. Thanks!

alexmohr:
Works for me with default cmake arguments thank you. Now that it works with clang we could update the brew file too, so it does not depend on gcc anymore?

minminlittleshrimp:
IMHO we still support both, and in Apple case here clang still be fine right? homebrew with prevelent using gcc can also alternatively with clang?

I have check with local Clang and this change is valid, and I believe it good to go with the configuring.

Im also deploying 3 more pipelines for clang tidy clang format cppchecker and commit msg format checker s.a. requesting signoff or limit to 72 chars soem kind of, that can help strengthening the automation process

alexmohr:
Yes we still support both with this PR, but as @felipek also mentioned already clang is pretty much the standard way on MacOS, so we could drop the gcc depedency from the brew file and compile it with clang per default. This will only affect users installing via homebrew and you still can build locally with GCC instead.

felipek:
Hi @minminlittleshrimp @alexmohr yes clang is the Apple standard (llvm/clang are Apple originated projects anyway), gcc requires either brew/MacPorts or something similar. I think it is fine to keep support for both, but it is more testing/coverage, yes...

minminlittleshrimp:
Thanks, this patch is nicely confirming the info I need, shall go on with this patch as followup PR.

By the way just my curiousity: why we need dlt-daemon on MacOS? I would believe there some specific projects or purpose here? 
😅

minminlittleshrimp (approved):

minminlittleshrimp (changes requested):
From me: I shall work on rebasing this, no worry, this must be the next PR we taking care after guardrail apply 😀

alexmohr:
> Thanks, this patch is nicely confirming the info I need, shall go on with this patch as followup PR.
> 
> By the way just my curiousity: why we need dlt-daemon on MacOS? I would believe there some specific projects or purpose here? 😅

I can at least answer that partially / for my use case. 
MacOS is my daily driver at the moment and when working on projects that use dlt it comes in handy when you have all dependencies on your local machine, so you don't have to deal with docker or virtual machines and just can build everything natively. It also helps when you have tools like 'dlt-receive' or 'dlt-control' available on a local machine when we're testing in a vehicle, because it makes the network setup easier. Although the dlt-viewer is available in the appstore, I do prefer the command line tools.

minminlittleshrimp (approved):
Look good to me

PR #905, opened 2026-08-06
Create MSYS2.yaml and fix some issues on MSYS2.

minminlittleshrimp:
Add MSYS2 building action to check building issues on MSYS2.

Fix some "error=char-subscripts" for isdigit and isspace functions by adding explicit cast to int,
which is required by C99 standard and later.


minminlittleshrimp:
Hello @pffang please answer why we set baudrate to unused? I believe it use in switch case block?

pffang:
> Hello @pffang please answer why we set baudrate to unused? I believe it use in switch case block?

You can check the macro detection inside the speed_t dlt_convert_serial_speed(int baudrate) function. MSYS will define the __CYGWIN_ macro, so it will go to the #else branch, and this branch doesn't use the baudrate parameter.

minminlittleshrimp:
@pffang This PR somehow block gtest dlt user from finishing, please check the touching in dlt common to see if the root cause from there

pffang:
> @pffang This PR somehow block gtest dlt user from finishing, please check the touching in dlt common to see if the root cause from there

That's strange. My changes only adapt to the strict grammar checking in the newer compiler, don't include any logic modification.

minminlittleshrimp (commented):
Look good to me

minminlittleshrimp:
Hi @santhoshsivanhere 
Kindly review and approve -> squash and merge
Thanks

src/shared/dlt_common.c
pffang, line 1: You need to keep `DLT_UNUSED(baudrate);` brefore `return 0;`.
Or it will not compiled on MSYS2.

src/daemon/dlt-daemon.c
```diff
@@ -535,364 +511,240 @@ int option_file_parser(DltDaemonLocal *daemon_local)
...
+                    } else if (strcmp(token, "OfflineLogstorageDelimiter") == 0) {
                         /* Check if valid punctuation, default otherwise*/
-                        if (ispunct((int)value[0]))
-                            daemon_local->flags.offlineLogstorageDelimiter = value[0];
-                    }
-                    else if (strcmp(token, "OfflineLogstorageMaxCounter") == 0)
-                    {
-                        daemon_local->flags.offlineLogstorageMaxCounter = (unsigned int) atoi(value);
-                        daemon_local->flags.offlineLogstorageMaxCounterIdx = (unsigned int) strlen(value);
+                        if (ispunct((char)value[0]))
```
pffang, line 659: Can you use `ispunct((int)value[0]` ? The function prototype for ispunct takes an int as a parameter. You can check the C library manual for this.

src/daemon/dlt-daemon.c
```diff
@@ -2866,79 +2585,74 @@ int dlt_daemon_log_internal(DltDaemon *daemon, DltDaemonLocal *daemon_local,
...
+int dlt_daemon_check_numeric_setting(char* token, char* value, unsigned long* data)
 {
     char value_check[value_length];
     value_check[0] = 0;
     sscanf(value, "%lu%s", data, value_check);
-    if (value_check[0] || !isdigit((int)value[0])) {
-        fprintf(stderr, "Invalid input [%s] detected in option %s\n",
-                value,
-                token);
+    if (value_check[0] || !isdigit(value[0])) {
```
pffang, line 2617: Can you use `isdigit((int)value[0]` ? Like `ispunct` The function prototype for `isdigit` also takes an int as a parameter. You can check the C library manual for this.

pffang (changes requested):

pffang (changes requested):
A few places still need to be changed to pass the MSYS2 building.

minminlittleshrimp:
https://github.com/COVESA/dlt-daemon/pull/913/commits

A fully rebase-nicely with drop-in fix has been created

Issue #551, opened 2023-10-19
DLT-Multinode: Gateway does not recognize reset of passive node


marc-heinlein:
**Precondition:**
1. Configure a multinode setup like described [here](https://github.com/COVESA/dlt-daemon/blob/master/doc/dlt_multinode.md)
2. Passive and Gateway node are up and running, Gateway has connected to Passive Node, DLT logs of Passive Node can be received at client connected to GatewayNode

**Action:**
- Reset Gateway Node without proper shutdown (i. e. do not stop dlt-daemon)

**Expected behaviour:**
- Gateway Node recognizes unavailability of Passive node and starts reconnecting

**Observed behavior:**
- Gateway Node does not recognize unavailability of Passive node and no reconnection attempt is started

Traces with Verbose=1 and LoggingLevel=7 are attached: [dlt-log.zip](https://github.com/COVESA/dlt-daemon/files/13044341/dlt-log.zip)

DLT Package Version: 2.18.9 STABLE, Package Revision: v2.18.9

marc-heinlein:
> Might be solved here: #537
> 
> I will test it...

This does not solve the issue.

michael-methner:
Hello @marc-heinlein ,
thanks for your report.
I have one questions:
"Reset Gateway Node without proper shutdown (i. e. do not stop dlt-daemon)"
Do you mean reset of Gateway Node or Passive Node here?
I would assume on reset on Gateway Node, it would try reconnecting after reboot (as dlt-daemon will be starting again).

marc-heinlein:
Hello @michael-methner, the issue is when resetting the passive node. When resetting the gateway node, everything works like expected.

thanhbnq:
Hi all,
As I remember, this is an expected behavior.
The connection between gateway and passive nodes are established at starting.
Gateway defines some retrials at beginning and timeout there to connect to passive.
After that, there won't be any effort to reconnect passive nodes.

You can try to trigger an establishment from passive via dlt-passive-node-ctrl.

marc-heinlein:
Hmm, for me, it sounds strange, that the passive node should actively request at the gateway node to reconnect to passive node. If this is the forseen strategy for the multinode configuration, it should be built into the passive node, that it automatically (re-) requests the connection once the gateway node does not connect until a certain timeout (i. e. if the passive node restarts after a reset).

[...]

PR #912, opened 2026-08-25
Fix DLT-Multinode: Gateway does not recognize reset of passive node #551

minminlittleshrimp:
Enable TCP keepalive to detect broken connections due to network disconnects. Without this in cases where no TCP FIN or RST package is received from the server, the dlt_client will never notice that it was disconnected from the server. Thus it will not reconnect when the server is available again. Also improve logging output to see which connection fails.


minminlittleshrimp:
This fix require a setup of 2 daemons connection (docker + host) for last testing/validation

minminlittleshrimp:
rebase and rework after clang-merged: https://github.com/COVESA/dlt-daemon/pull/914

PR #913, opened 2026-09-13
[Prior 2] Create MSYS2.yaml and fix some issues on MSYS2.

minminlittleshrimp:
Add MSYS2 building action to check building issues on MSYS2.

Fix some "error=char-subscripts" for isdigit and isspace functions by adding explicit cast to int,
which is required by C99 standard and later.

PR #915, opened 2026-09-13
[Prior 1] dlt_user: fix deadlock by stopping threads before locking dlt_mutex

minminlittleshrimp:
dlt_user: fix deadlock by stopping threads before locking dlt_mutex
dlt_free() held dlt_mutex while calling dlt_stop_threads(), but the
housekeeper thread needs dlt_mutex to proceed (e.g. inside
dlt_user_log_check_user_message). Since pthread_mutex_lock is not a
cancellation point, pthread_cancel could not interrupt the blocked
housekeeper, causing pthread_join to deadlock.

- Move dlt_stop_threads() before dlt_mutex_lock() in dlt_free()
- Add cooperative exit check at top of housekeeper loop (~500ms latency)
- On MSYS2/MinGW: cooperative wait (1s) + pthread_cancel fallback
- On Linux/other: pthread_cancel directly (no added latency)
- Reset dlt_user_housekeeper_exit_requested in dlt_start_threads()
  to prevent stale flag from previous dlt_free() cycle
- Guard CLOCK_MONOTONIC for MSYS2/MinGW in condattr and clock_gettime
- Add pthread_condattr_destroy() after pthread_cond_init()


minminlittleshrimp:
Hello @santhoshsivanhere 
here we go again, leaks show up everywhere :)) https://github.com/COVESA/dlt-daemon/actions/runs/34745962782/job/103693843991?pr=915
Please review this PR, and as it shows leaks, I propose to fix them in this PR also, what do ya think?

minminlittleshrimp:
Discussion has been made, split this into 3 PRs.
- Fix deadlock libdlt
- Fix memleak for also protocol v2 APIs
- Introduce sanitizer pipe

src/lib/dlt_user.c
```diff
@@ -7114,9 +7148,54 @@ void dlt_stop_threads()
...
+            if (pthread_kill(dlt_housekeeperthread_handle, 0) != 0) {
+                /* Thread has already terminated */
+                break;
+            }
+            ts.tv_sec = 0;
+            ts.tv_nsec = DLT_USER_RECEIVE_NDELAY;
+            nanosleep(&ts, NULL);
+        }
+
+        if (pthread_kill(dlt_housekeeperthread_handle, 0) != 0) {
```
santhoshsivanhere: Hey @minminlittleshrimp, Please check if below AI code review comment a valid concern?

`pthread_kill(thread, 0)` is not a reliable portable test for whether a pthread has returned but has not yet been joined. On implementations where the `pthread_t` remains valid until `pthread_join()`, it can continue returning success after the thread has cooperatively exited. That means the code can still execute the `pthread_cancel()` fallback even though the thread has already finished.

This undermines the stated MSYS2/MinGW goal and may reintroduce the platform-specific cancellation behavior the workaround is intended to avoid.

**Suggested fix:** use an explicit atomic “thread exited” state, or signal a condition variable from the housekeeper immediately before returning and wait on that condition with a bounded timeout. Then call `pthread_join()` unconditionally; use cancellation only after the bounded cooperative wait expires.

santhoshsivanhere: The comment is valid. This will be handled in future. This is being merged to unblock PR #913

santhoshsivanhere (approved):

PR #917, opened 2026-09-21
Fix resource leaks, signal handling, and clean lifecycle management in dlt-qnx-system

### LINKED ISSUE #890: dlt-qnx-slogger2-adapter: resource leaks and robustness issues in context-map parsing and cleanup
### Environment

- `src/dlt-qnx-system/dlt-qnx-slogger2-adapter.cpp` (and `dlt-qnx-system.h`)
- `master`, project version 3.0.1

I was reading through the slogger2 adapter and ran into a handful of issues. A couple are clear bugs, the rest is ownership and cleanup stuff that's easy to get wrong as it stands. Listing them together; happy to split if you'd rather.

### Bugs

The JSON decoder in `dlt_context_map_read` is never freed. `json_decoder_create()` allocates it but there's no `json_decoder_destroy()` anywhere, not on success, not on the early `return` after a parse failure, not on the `break` paths. So it leaks on every call ([json_decoder_destroy docs](https://www.qnx.com/developers/docs/7.0.0/com.qnx.doc.json/topic/json_decoder_destroy.html)).

In the same loop, a single bad entry kills the whole map. If descending into a context object or reading its `name`/`description` fails, the code `break`s and stops parsing everything after it, so one malformed entry silently drops all the valid contexts that follow. I think it should pop + continue and skip just the bad one.

`wait_for_buffer_space` divides by `total_size` without checking it's non-zero (`used_size / total_size`). If `dlt_user_check_buffer` ever returns a zero total that's UB. Probably want to treat zero as "full" and bail.

`clean_qnx_slogger2` deletes every `DltContext*` in `g_slog2file` without unregistering it first. The adapter's own `dltQnxSlogger2Context` does get a `DLT_UNREGISTER_CONTEXT` (in `slogger2_thread`), but the per-name contexts registered from the JSON map via `dlt_register_context` are only `delete`d, never unregistered, and the map isn't cleared. That can leave libdlt with dangling registrations pointing at freed memory.

### Ownership / robustness

Most of the above comes back to the map being `unordered_map<string, DltContext*>` with raw `new`/`delete`. Switching the value to `unique_ptr<DltContext>` and giving the json decoder an RAII guard would make the leaks and the manual cleanup loop go away on their own. Related: in the parse loop, `new DltContext` then `emplace(...)` leaks the context if `emplace` throws.

One more: `start_qnx_slogger2` grabs the thread stack with `malloc(PTHREAD_STACK_4K * 4)` and aligns it by hand (there's even a comment proving the result stays in bounds). `posix_memalign` or just `pthread_attr_setstacksize` would drop all of that.

### Minor

Diagnostics are mixed. Setup and error paths use `printf`/`fprintf(stderr)` while the rest of the file goes through `DLT_LOG`, and on a QNX target the stderr ones may not show up. There's also the usual const-correctness and C-style cast cleanup, plus `MALLOC_ASSERT` in the header doesn't parenthesize its argument (`if(x == NULL)`), which breaks for non-trivial expressions.

### Note + question

Docs mention a single-threaded redesign of `dlt-qnx-system` since 2.18.10, but the adapter on `master` still has the separate `slogger2_thread` and the manual stack setup, so if a rewrite is already in progress somewhere, I'd rather coordinate than duplicate it.

Otherwise I'd like to clean this up around RAII ownership, fix the parse-loop and cleanup logic, and simplify the stack handling. Would you prefer I keep it to the clear bugs first, or take the broader refactor? And is there history behind the current design I should be aware of before touching it? Happy to do it as small compilable commits per the contribution guide.

santhoshsivanhere:
## Summary

This PR hardens the `dlt-qnx-system` slogger2 adapter against resource
leaks, undefined behavior, and unclean shutdowns. It replaces heap-based
thread stack allocation with `mmap`, introduces RAII-based ownership for
DLT contexts and JSON decoders, and reworks signal handling so the daemon
terminates synchronously and predictably.

### 1. Resource leaks and exception safety in slogger2
- Wrap `json_decoder_t` with an RAII deleter to guarantee cleanup on all
  exit paths in `dlt_context_map_read()`.
- Continue parsing remaining entries when JSON object/key extraction fails
  instead of breaking early.
- Guard against `total_size <= 0` in `wait_for_buffer_space()` to prevent
  division by zero.
- Explicitly unregister all `DltContext` entries in `clean_qnx_slogger2()`
  before tearing down storage.
- Replace raw pointers in `g_slog2file` with `std::unique_ptr<DltContext>`
  to enforce exception-safe ownership.

### 2. Replace malloc-based thread stack allocation with mmap
- Replace `malloc()` / `free()` for the custom thread stack with `mmap()`
  (`MAP_ANONYMOUS | MAP_PRIVATE`) and `munmap()`.
- Introduce `STACK_ALLOC_SIZE` so allocation and cleanup use identical
  bounds.
- Check against `MAP_FAILED` for accurate error handling.
- Removes the need for manual page-alignment calculations before calling
  `pthread_attr_setstack()`.

### 3. Clean lifecycle management and signal handling
- Block `SIGINT`, `SIGTERM`, `SIGHUP`, `SIGQUIT`, and `SIGALRM` via
  `pthread_sigmask()` *before* thread creation.
- Synchronously wait for termination signals using `sigwait()` to avoid
  async signal-handler deadlocks.
- Add atomic flag `g_slog2_thread_alive`, checked inside
  `slogger2_callback()` and `wait_for_buffer_space()`, to break out early
  on shutdown.
- Ensure the `mmap`-allocated thread stack is cleanly unmapped.
- Iterate through `g_slog2file` to unregister all dynamically registered
  DLT contexts on thread exit.
- Guarantee `pthread_join()` completes before calling `clean_up()` and
  unregistering the main DLT application.

## Motivation

The slogger2 adapter could leak JSON decoders, dereference dangling
`DltContext` pointers during teardown, divide by zero on empty buffers,
and terminate uncleanly on signals. Together these fixes make the
lifecycle of the adapter deterministic and exception-safe.

## Testing

- Built `dlt-daemon` with `-DWITH_DLT_QNX_SYSTEM=ON`.
- Verified the daemon starts, registers contexts, and shuts down cleanly
  on `SIGTERM` / `SIGINT`.
- Confirmed no leaks reported under the previous code paths.

Fixes #890.

Signed-off-by: Santhosh Sivan Murugan <Murugan.SanthoshSivan@in.bosch.com>


santhoshsivanhere:
@minminlittleshrimp Please spend some time to review this PR to fix issue #890.

src/dlt-qnx-system/dlt-qnx-slogger2-adapter.cpp
```diff
@@ -50,104 +52,122 @@ std::atomic<bool> g_slog2_thread_alive(false);
 extern DltContext dltQnxSystem;
 
 static DltContext dltQnxSlogger2Context;
-
 static std::set<std::string> dltWarnedMissingMappings;
 
 extern DltQnxSystemThreads g_threads;
 
-static std::unordered_map<std::string, DltContext*> g_slog2file;
+static std::unordered_map<std::string, std::unique_ptr<DltContext>> g_slog2file;
```
minminlittleshrimp, line 63: DltContext is a C struct, so it still need dlt native api lifecycle (dlt_register_context/unregis context), we can do like this:
```
struct DltContextDeleter {
    void operator()(DltContext *ctx) const {
        if (ctx) {
            dlt_unregister_context(ctx);
            delete ctx;
        }
    }
};
using DltContextPtr = std::unique_ptr<DltContext, DltContextDeleter>;
```

Then:
```
static std::unordered_map<std::string, DltContextPtr> g_slog2file;
```

in cleanup simply:
```
void clean_qnx_slogger2()
{
    g_slog2file.clear();  // we ok here
    dltWarnedMissingMappings.clear();
    free_stackaddr();
}
```
just RAII same as Jsondecoderdeleter, what do you think?
Thanks

src/dlt-qnx-system/dlt-qnx-slogger2-adapter.cpp
```diff
@@ -239,41 +268,36 @@ static int slogger2_callback(slog2_packet_info_t *info, void *payload, void *par
...
-    ret = dlt_user_log_write_start(ctxt, &log_local, loglevel);
+    int ret = dlt_user_log_write_start(ctxt, &log_local, loglevel);
 
-    /* OK means loglevel under threshold */
     if (ret == DLT_RETURN_OK) {
         return 0;
     }
 
     if (ret != DLT_RETURN_TRUE) {
-        fprintf(stderr, "%s: could not log to DLT status=%d\n", __func__, ret);
```
minminlittleshrimp, line 259: Why remove this? thanks

src/dlt-qnx-system/dlt-qnx-system.c
```diff
@@ -415,14 +362,15 @@ static void start_thread()
     }
 }
 
-/**
- * Wait for threads to exit.
- */
-static void join_thread()
+/* Point 4: Correct Joining & Context Unregister Sequence */
```
minminlittleshrimp: what it means by point 4? thanks

src/dlt-qnx-system/dlt-qnx-slogger2-adapter.cpp
```diff
@@ -296,23 +319,18 @@ static void *slogger2_thread(void *v_conf)
...
-                    "slog2_parse_all() stops working.\n");
+                    "slog2_parse_all() stopped processing.\n");
     }
 
-    DLT_LOG_CXX(dltQnxSystem, DLT_LOG_DEBUG, __func__, ": Exited main loop.");
+    DLT_LOG_CXX(dltQnxSystem, DLT_LOG_DEBUG, __func__, ": Exited ring buffer loop.");
 
     DLT_UNREGISTER_CONTEXT(dltQnxSlogger2Context);
-    /* Try to send SIGTERM to make sure main thread wakes up for cleaning */
-    pthread_kill(g_threads.main_thread, SIGTERM);
```
minminlittleshrimp, line 315: remove pthread_kill here with no wakeup mechanism can hang on sigwait forever? A heartbeat needed, are we missing dlt_slog2_heartbeat_loop? Thanks

src/dlt-qnx-system/dlt-qnx-system.c
```diff
@@ -131,60 +124,45 @@ int main(int argc, char* argv[])
...
-/**
- * Initialize command line options with default values.
- */
 static void init_cli_options(DltQnxSystemCliOptions *options)
 {
     options->configurationFileName     = DEFAULT_CONF_FILE;
     options->daemonize                 = 0;
 }
 
-/**
```
minminlittleshrimp, line 185: why remove brief/doxygen/comments/license/unrelated comments? Thanks

src/dlt-qnx-system/dlt-qnx-slogger2-adapter.cpp
santhoshsivanhere: I agree

